PR #18, opened 2025-05-05
plugins: expose Plugin interface publicly

### LINKED ISSUE #15: Expose plugin.go interface from verifier
With the repo split, we need to re-introduce the plugin.go interface in the verifier repo - this must then be consumed by the plugin implementations

RaghavSood:
Resolved #15

This brings in the plugin interface to the verifier.

We want this to be a top-level publicly accessible interface as it defines the exact functionality that must be implemented by a package to be considered a plugin.

Anyone who implements these (the list will grow) will be more or less ready to be a plugin for vultisig - they will still have to make a PR to the verifier to implement some verification logic and register a plugin ID, but the core functionality is encapsulated in this interface.

I have removed the `embed.FS` option from the previous repo since that is no longer required with the planned shift to a config-schema based UI

## Summary by CodeRabbit

- **New Features**
  - Introduced a new plugin interface to support customizable plugin behaviors, including policy validation, transaction proposal, transaction validation, and handling of completed signing operations.

### DIFF
--- a/plugin/plugin.go
+++ b/plugin/plugin.go
@@ -1,0 +1,15 @@
+package plugin
+
+import (
+	"context"
+
+	"github.com/vultisig/mobile-tss-lib/tss"
+	"github.com/vultisig/verifier/types"
+)
+
+type Plugin interface {
+	ValidatePluginPolicy(policyDoc types.PluginPolicy) error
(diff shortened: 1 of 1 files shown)


johnnyluo (approved):
LGTM

PR #22, opened 2025-05-05
address: return public key data and support EdDSA derivations through public entrypoint

### LINKED ISSUE #21: common.GetAddress should return address, publicKey, and curve type
Presently, common.GetAddress returns only the address

However, we should return the public key as a hex string, and the curve type (following the `isEdDSA` convention in the rest of the codebase) - this is necessary as some flows, such as policy creation, require signature verification.

For some chains, the public key is hashed during the formulation of the address - while it is possible to resort to flows such as ecrecover, it can be helpful in some cases to have the public key available alongside the address.

RaghavSood:
Resolves #21 

This fixes two gaps:

1. It returns the derived public key for ecdsa chains, along with a bool flag indicating the curve type
2. It only attempts to derive subkeys for ecdsa chains - for EdDSA it uses the root key - previously, address.GetAddress would fail when called for EdDSA chains. This fixes that and adds a test case to cover regression

## Summary by CodeRabbit

- **Refactor**
  - Improved address generation to include the public key and EdDSA status along with the address.
- **Tests**
  - Expanded tests to cover address and public key generation for standard and EdDSA keys, adding new cases for the Sui chain and validation of error handling for invalid inputs.

### DIFF
--- a/address/address.go
+++ b/address/address.go
@@ -8,42 +8,61 @@ import (
 	"github.com/vultisig/verifier/common"
 )
 
-// GetAddress returns the address for the given public key and chain.
-func GetAddress(rootHexPublicKey string, rootChainCode string, chain common.Chain) (string, error) {
-	hexPublicKey, err := tss.GetDerivedPubKey(rootHexPublicKey, rootChainCode, chain.GetDerivePath(), chain.IsEdDSA())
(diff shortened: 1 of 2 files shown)


johnnyluo (approved):
LGTM

PR #40, opened 2025-05-08
policy: remove DerivePath as top level field

### LINKED ISSUE #20: Remove DerivePath from PluginPolicy
Policies are not chain specific and should not tie themselves to a specific derivation path.

With the completion of https://github.com/vultisig/verifier/pull/19 we should be able to do away with ad-hoc derivation paths and strictly use the pre-determined ones when the policy calls for it.

We can only do this on verifier after https://github.com/vultisig/plugin/issues/29 is implemented on plugins

RaghavSood:
Fixed #20

This is the verifier follow up to https://github.com/vultisig/plugin/pull/32 - we remove DerivePath from the top-level policy document as each individual action inside a policy should be defining its own derivation path (whether explicitly or implicitly - this is for a subsequent PR)

This also removes the corresponding column from the database.

## Summary by CodeRabbit

- **Refactor**
  - Removed the use of the derive path parameter from signature verification and related internal logic.
  - Updated the PluginPolicy structure to no longer include the derive path field.

- **Chores**
  - Migrated the database schema to remove the derive path column from plugin policies.

### DIFF
--- a/internal/api/policy.go
+++ b/internal/api/policy.go
@@ -65,7 +65,7 @@ func (s *Server) verifyPolicySignature(policy types.PluginPolicy, update bool) b
 		return false
 	}
 
-	isVerified, err := sigutil.VerifySignature(policy.PublicKey, policy.ChainCodeHex, policy.DerivePath, msgBytes, signatureBytes)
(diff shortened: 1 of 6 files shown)


johnnyluo (approved):
LGTM

PR #44, opened 2025-05-10
 Make verifier host configurable in worker service

SunshineRider:
Currently, the worker service assumes the verifier is running on localhost. This change makes the verifier host configurable through the config file, allowing the verifier to be deployed on any machine.

Changes:
- Added VerifierHost field to Server config struct
- Updated config.example.json with verifier_host field
- Modified initiateTxSignWithVerifier to use configured host with localhost fallback

This change enables:
- Deploying verifier on different machines / on different networks
- Maintaining backward compatibility with localhost fallback

## Summary by CodeRabbit

- **New Features**
  - Added new server configuration options, including customizable host, port, verifier host, and JWT secret in the configuration file.
  - The verifier host used for certain operations can now be set via configuration instead of being hardcoded.
- **Chores**
  - Updated configuration structure to support the new server and verifier host settings.

### DIFF
--- a/config.example.json
+++ b/config.example.json
@@ -22,5 +22,11 @@
   "datadog": {
     "host": "localhost",
     "port": "8125"
+  },
+  "server": {
+    "host": "localhost",
+    "port": 8080,
+    "verifier_host": "localhost",
+    "jwt_secret": "your-jwt-secret"
   }
 }
--- a/config/config.go
+++ b/config/config.go
@@ -28,9 +28,10 @@ type Config struct {
 
 type VerifierConfig struct {
 	Server struct {
-		Host     string `mapstructure:"host" json:"host,omitempty"`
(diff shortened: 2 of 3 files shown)


johnnyluo (commented):
I think this PR misunderstand the role of verifier worker 

verifier worker is not going to make RPC call back to verifier server 

Keysign request will be initiate from the plugin side , as far as I understand , verifier will not initiate keysign

SunshineRider:
> I think this PR misunderstand the role of verifier worker
> 
> verifier worker is not going to make RPC call back to verifier server
> 
> Keysign request will be initiate from the plugin side , as far as I understand , verifier will not initiate keysign

ah yes, closing this PR. here's the correct PR - https://github.com/vultisig/plugin/pull/40

PR #45, opened 2025-05-11
plugin/registry: introduce plugin registry with constants for DCA and payroll

### LINKED ISSUE #16: verifier: introduce plugin type registry
With the repo split, we've lost the plugin type registry - we need to have a verifier dictated, unique plugin type identifier for all plugins that can be used during flows such as policy validation and app store management/reshare set up.

We should bring this back at a top level under the `plugins` package (alongside the policy document), and require any new plugins to request a plugin ID as they onboard

RaghavSood:
Fixed #16

No functional changes, just introduces public constants for payroll and DCA

Subsequent PRs will use these in the policy refactors and recipe verification implementation flows

## Summary by CodeRabbit

- **New Features**
  - Introduced unique identifiers for plugins, ensuring each plugin can be distinctly recognized within the application.

### DIFF
--- a/plugin/registry.go
+++ b/plugin/registry.go
@@ -1,0 +1,13 @@
+package plugin
+
+type PluginID string
+
+// Each plugin must reserve a unique ID.
+//
+// A plugin ID is comprised of a all-lowercase string, ending with 4 random hex digits.
+//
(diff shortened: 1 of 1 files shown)


johnnyluo (approved):
LGTM

PR #68, opened 2025-05-15
goose: allow out of order migrations

### LINKED ISSUE #67: Goose should allow out of order migrations
Goose by default does not allow migrations to be run out of order - this results in errors when branches are long lived, and migrations with future timestamps have already been merged and applied

Goose has an option to support this - we should use it

RaghavSood:
Fixes #67 

This also fixes the server deployment which is presently stuck on pending migrations


johnnyluo (approved):
LGTM

PR #86, opened 2025-05-25
api: restore plugin signing API

RaghavSood:
Restores the plugin signing API from the combined repo

This isn't a 1:1 copy - I'm skipping some of the tx db handling since we have yet to refactor the verifier tx indexing, and the previous approach is no longer compatible with our updated db.

It also makes an assumption on party names that may or may not hold once we get reshare working.

This is required for the stacked PR that implements recipes for payroll policy verification.

### DIFF
--- a/internal/api/plugin.go
+++ b/internal/api/plugin.go
@@ -1,19 +1,84 @@
 package api
 
 import (
+	"encoding/json"
 	"errors"
 	"fmt"
 	"net/http"
 	"strconv"
+	"time"
 
+	"github.com/google/uuid"
+	"github.com/hibiken/asynq"
 	"github.com/jackc/pgx/v5"
 	"github.com/labstack/echo/v4"
 	"github.com/microcosm-cc/bluemonday"
 	"github.com/vultisig/verifier/common"
 
+	"github.com/vultisig/verifier/internal/tasks"
 	"github.com/vultisig/verifier/internal/types"
 	ptypes "github.com/vultisig/verifier/types"
 )
+
+func (s *Server) SignPluginMessages(c echo.Context) error {
+	s.logger.Debug("PLUGIN SERVER: SIGN MESSAGES")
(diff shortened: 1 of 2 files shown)


internal/api/plugin.go
```diff
@@ -1,20 +1,85 @@
...
+	result, err := s.redis.Get(c.Request().Context(), req.SessionID)
+	if err == nil && result != "" {
+		return c.NoContent(http.StatusOK)
+	}
+
+	if err := s.redis.Set(c.Request().Context(), req.SessionID, req.SessionID, 30*time.Minute); err != nil {
+		s.logger.Errorf("fail to set session, err: %v", err)
+	}
+
+	req.Parties = []string{"verifier", req.PluginID}
```
johnnyluo, line 62: req.parties should be the parties that will be joining the keysign , it is decided by the client,  we should not change it

internal/api/server.go
```diff
@@ -132,6 +132,7 @@ func (s *Server) StartServer() error {
 	pluginGroup := e.Group("/plugin", s.userAuthMiddleware)
 	pluginGroup.POST("/policy", s.CreatePluginPolicy)
 	pluginGroup.PUT("/policy", s.UpdatePluginPolicyById)
+	pluginGroup.POST("/sign", s.SignPluginMessages)
```
johnnyluo, line 135: There is a method call `SignMessages` currently we have , which is mostly inherit from vultiserver
As far as I understand , we will not need it 
You have two choice here, either keep your new `SignPluginMessages` , or update the `SignMessage` to behave as SignPluginMessage , we don't need both

RaghavSood: Yes, already planning to remove it this week, tracked in https://github.com/vultisig/verifier/issues/89

johnnyluo (approved):
LGTM

PR #87, opened 2025-05-25
payroll: use recipes for verification of tx

### LINKED ISSUE #25: Payroll policy verification
Implement payroll-specific policy verification

Depends on https://github.com/vultisig/verifier/issues/24

RaghavSood:
Stacked PR on https://github.com/vultisig/verifier/pull/86

Fixes #25

This works, but shouldn't be a final implementation

While doing this I realized we need to restructure the plugin keysign message a little to encode the network for each transaction (so we can build the correct matcher, it just assumes ETH for now), and may want to move the actual verification to be done in the worker.

However, this implements enough that we should be able to get a test transaction through once plugin server is ready to propose them.

## Summary by CodeRabbit

- **New Features**
  - Introduced policy-based validation for signing Ethereum transactions, ensuring only transactions that match defined policy rules are allowed for signing.

- **Chores**
  - Updated and cleaned up dependencies, including upgrades and additions to cryptographic and utility libraries.


johnnyluo (approved):
LGTM

PR #146, opened 2025-06-09
Adding verifier call for reshare 

SunshineRider:
## Summary by CodeRabbit

- **New Features**
  - Added vault reshare capability with plugin server notifications and backend integration.
- **Improvements**
  - Enhanced error handling, logging, and user feedback during the vault reshare process.
  - Updated plugin installation flow to support reshare sessions with plugin-specific context.
- **Bug Fixes**
  - Strengthened request validation and processing for more reliable reshare operations.

### DIFF
--- a/internal/api/server.go
+++ b/internal/api/server.go
@@ -243,7 +243,6 @@ func (s *Server) ReshareVault(c echo.Context) error {
 
 	// Store session in Redis
 	if err := s.redis.Set(c.Request().Context(), req.SessionID, req.SessionID, 5*time.Minute); err != nil {
-		s.logger.Errorf("fail to set session, err: %v", err)
 		s.logger.Errorf("ReshareVault: Failed to store session in Redis: %v", err)
 	}
 
@@ -266,6 +265,12 @@ func (s *Server) ReshareVault(c echo.Context) error {
 
 	// Notify plugin server asynchronously
 	go func() {
+		defer func() {
+			if r := recover(); r != nil {
+				s.logger.Errorf("ReshareVault: Panic in async plugin notification: %v", r)
+			}
+		}()
+
 		bgCtx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
 		defer cancel()
 
--- a/internal/api/server.go
+++ b/internal/api/server.go
@@ -213,8 +213,6 @@ func (s *Server) ReshareVault(c echo.Context) error {
 	s.logger.Info("ReshareVault: Starting reshare vault request")
 
 	var req tv.ReshareRequest
-
-	// Parse and validate request
 	if err := c.Bind(&req); err != nil {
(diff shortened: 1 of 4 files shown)


internal/api/server.go
```diff
@@ -232,11 +259,71 @@ func (s *Server) ReshareVault(c echo.Context) error {
 		asynq.Retention(10*time.Minute),
 		asynq.Queue(tasks.QUEUE_NAME))
 	if err != nil {
+		s.logger.Errorf("ReshareVault: Failed to enqueue task: %v", err)
 		return fmt.Errorf("fail to enqueue task, err: %w", err)
 	}
+
+	// Notify plugin server asynchronously
+	go func() {
```
johnnyluo, line 267: @SunshineRider , please don't make request to Plugin server from verifier , the front-end ui part should call the plugin server for this

internal/api/server.go
```diff
@@ -232,11 +259,71 @@ func (s *Server) ReshareVault(c echo.Context) error {
...
+		if err := s.notifyPluginServerReshare(bgCtx, req); err != nil {
+			s.logger.Errorf("ReshareVault: Failed to notify plugin server: %v", err)
+		}
+	}()
+
 	return c.NoContent(http.StatusOK)
 }
 
+// notifyPluginServerReshare sends the reshare request to the plugin server
+func (s *Server) notifyPluginServerReshare(ctx context.Context, req tv.ReshareRequest) error {
```
johnnyluo, line 286: Do this from the front end

johnnyluo (changes requested):
See comments

internal/api/server.go
SunshineRider: but should we expose this function for UI to use or directly call the plugin server from FE?

internal/api/server.go
```diff
@@ -232,11 +249,71 @@ func (s *Server) ReshareVault(c echo.Context) error {
 		asynq.Retention(10*time.Minute),
 		asynq.Queue(tasks.QUEUE_NAME))
 	if err != nil {
+		s.logger.Errorf("ReshareVault: Failed to enqueue task: %v", err)
 		return fmt.Errorf("fail to enqueue task, err: %w", err)
 	}
+
+	// Notify plugin server asynchronously
+	go func() {
```
johnnyluo, line 257: instead of using a go i think it will be better we just make a RPC call directly.
1. RPC call to plugin server , if fail , directly return to client
2. If 1 succeed , then queue the task to redis , and trigger verifier to join 
3. and then return to client 

Currently , using a go routine , it is fire and forget , even Plugin is offline , the front end has no way to know

johnnyluo (approved):
LGTM

internal/api/server.go
SunshineRider: made the fix

PR #147, opened 2025-06-10
ui: fix auth

RaghavSood:
Fixes auth to work with the updated extension

1. Passes in the wallet address before sign in is complete - previously, before we had sign in working, we set the connected wallet data into the store immediately - now, it happens after auth, but the sign in message tries to reference the store, which results in an empty address - this should be fixed more properly to use the selected vault, not just default to the first one though
2. Adjusts the recovery ID to match go-ethereum's `crypto.Ecrecover` expectations - it requires the recovery ID to be 0-4, while extensions (not just vultisig) typically produce 27-31.

https://github.com/ethereum/go-ethereum/blob/3cc0e7a31ab8d070dbacf77412f935b6bb789054/crypto/secp256k1/secp256.go#L167-L169 

## Summary by CodeRabbit

- **New Features**
  - Enhanced development environment with TypeScript and JavaScript support, including Node.js 22 and npm integration.

- **Refactor**
  - Updated wallet authentication flow to explicitly use the wallet address during message signing.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -63,6 +63,17 @@
                     enable = true;
                   };
 
+                  languages.typescript = {
+                    enable = true;
+                  };
+
+                  languages.javascript = {
+                    enable = true;
+                    package = pkgs.nodejs_22;
+                    npm.enable = true;
+                  };
+
+
                   packages = with pkgs; [ 
                     redis
                     goose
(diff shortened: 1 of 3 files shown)


johnnyluo (approved):
LGTM

PR #176, opened 2025-06-18
Add connection check for install plugin

### LINKED ISSUE #169: [UI] Install plugin button should only be enabled after sign in is complete
The button to install a plugin should only be enabled once the user signs in

When signed out, it should be disabled/replaced with a message asking the user to install and connect the extension first

Image

Ehsan-saradar:
Fix #169 

## Summary by CodeRabbit

- **New Features**
  - The "Install" button now prompts users to connect their wallet if not already connected, updating its label to "Connect" accordingly.
  - Once connected, users can proceed with installation directly from the same button.

- **Improvements**
  - Enhanced error handling during the installation process for a smoother user experience.


johnnyluo (approved):
LGTM

PR #178, opened 2025-06-19
Fix connection to extension

Ehsan-saradar:
## Summary by CodeRabbit

- **Bug Fixes**
  - Improved wallet authentication by preventing redundant signing and token retrieval when already authenticated with the same public key.


johnnyluo (approved):
LGTM

PR #186, opened 2025-06-23
feat: implement DeleteAllPolicies method and integrate it into plugin…

johnnyluo:
… deletion process

## Summary by CodeRabbit

- **New Features**
  - Added the ability to delete plugins via a new DELETE endpoint, which also cleans up associated policies and vault data.
- **Improvements**
  - Enhanced middleware usage for improved route security and consistency.
- **Removals**
  - Deprecated transaction sync endpoints have been removed.
- **Bug Fixes**
  - Improved error handling and logging during plugin server notifications.
- **Tests**
  - Extended test mocks to support policy deletion functionality.


garry-sharp (approved):
As per previous comment.

johnnyluo:
> So I'm happy to approve this, however, this logic may need to change down the line. The way I have structured the fee stuff so far is 1 plugin -> several policies -> several billing records (from the fee recipe rules, start dates etc) -> fees.
> 
> For audit purposes I'd be more in favour of marking a plugin "inactive" and then having a dB rule that prevents any further update to FK'd records or itself.
> 
> That way people can still see a full list of transactions and fees accrued as well as other historical data. Again, this is merely my view as to what we may need in the future and happy to approve for now.

So this PR is to remove plugin policy that added to my vault , it is not to delete the plugin itself

PR #201, opened 2025-06-29
Add migration for transaction_history

Ehsan-saradar:
Add missing migration for transaction history

## Summary by CodeRabbit

* **New Features**
  * Added support for tracking transaction history, including status updates and error messages, for improved visibility into transaction lifecycles.


RaghavSood:
Don't think we should be merging this one, let's complete the clean up PR and get that in instead

PR #203, opened 2025-06-29
Add missing fields to Pricing

Ehsan-saradar:
## Summary by CodeRabbit

* **New Features**
  * Added creation and update timestamps to pricing information, allowing users to view when pricing entries were created and last updated.

### DIFF
--- a/internal/api/policy.go
+++ b/internal/api/policy.go
@@ -28,7 +28,7 @@ func (s *Server) validatePluginPolicy(policy types.PluginPolicy) error {
 		return errors.New("recipe cannot be empty")
 	}
 
-	recipeBytes, err := base64.RawStdEncoding.DecodeString(policy.Recipe)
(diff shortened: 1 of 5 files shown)


RaghavSood:
Hmmm, I'm not sure if this is something we want to be exposing.

I'm open to it, but thus so far pricings are shared across plugins - In the long run, this is perhaps unrealistic beyond obvious common pricings like FREE and $5 installation fee, etc.

I imagine subscription based plugins will mostly devolve into custom pricing that is per-plugin, so we could reasonably show some kind of timestamps here.

Do we have an existing design inclination that requires this? If not, I'd prefer to leave it out for the moment

Ehsan-saradar:
I did not change any design or flow, just trying to fix the current pricing endpoint issue (https://store.vultisigplugin.app/api/pricing/cbaa280f-24df-4e65-b524-fe7b6b5e6a1d) to make sure extension can interface with marketplace

PR #212, opened 2025-07-01
common: handle unencrypted vaults during parsing

### LINKED ISSUE #211: Handle unencrypted keyshare
Presently, when keyshares are sideloaded they may be exported unencrypted from an extension/mobile app. This results in vault handling failing

RaghavSood:
Fixes #211

We b64 encode vault data before storing it in the container. However, the current decryption flow in utils only decodes it during the decription process - if a vault is not encrypted, the b64 decoding is skipped, and we attempt to parse a b64 string directly as a protobuf message.

This aligns the decoding with other apps so that we can handle both cases

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of vault backups that are not encrypted, ensuring proper loading and decoding in both encrypted and unencrypted scenarios.

### DIFF
--- a/common/utils.go
+++ b/common/utils.go
@@ -151,6 +151,11 @@ func DecryptVaultFromBackup(password string, vaultBackupRaw []byte) (*vaultType.
 		if err != nil {
 			return nil, err
 		}
+	} else {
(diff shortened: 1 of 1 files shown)


webpiratt (approved):
LGTM

PR #233, opened 2025-07-07
🚧 Added hot reloading and reduced time on healthchecks to help docker performance

garry-sharp:
… performance

### DIFF
--- a/docker-compose.dev.yaml
+++ b/docker-compose.dev.yaml
@@ -1,0 +1,33 @@
+services:
+  verifier:
+    build:
+      context: .
+      dockerfile: Dockerfile.dev.verifier
+    volumes:
+      - ./cmd:/app/cmd
+      - ./internal:/app/internal
+      - ./address:/app/address
+      - ./common:/app/common
+      - ./config:/app/config
+      - ./plugin:/app/plugin
+      - ./testdata:/app/testdata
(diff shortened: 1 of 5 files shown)


docker-compose.dev.yaml
```diff
@@ -0,0 +1,35 @@
+services:
+  verifier:
+    build:
+      context: .
+      dockerfile: Dockerfile.dev.verifier
+    volumes:
+      - ./cmd:/app/cmd
```
johnnyluo, line 7: why not just map everything to the container like

`-  ./ : /app`

docker-compose.yaml
```diff
@@ -27,7 +27,7 @@ services:
       - shared_network
     healthcheck:
       test: [ "CMD", "redis-cli", "ping" ]
-      interval: 1s
+      interval: 10s
```
johnnyluo, line 30: why internal `10s`?  redis should be fast

docker-compose.dev.yaml
garry-sharp: Considered that, but only mounting the actual volumes here to keep the container isolated from the rest of the source tree. The idea is to avoid unintended rebuilds, it's easy when mounting the whole thing to also trigger infinite loops as it watches its own rebuild binaries and logs

PR #240, opened 2025-07-09
fix payroll data structures & policy validation

webpiratt:
Ref: https://github.com/vultisig/plugin/issues/124

### DIFF
--- a/internal/api/policy.go
+++ b/internal/api/policy.go
@@ -9,19 +9,17 @@ import (
 	"strconv"
 	"strings"
 
+	ecommon "github.com/ethereum/go-ethereum/common"
 	"github.com/google/uuid"
 	"github.com/labstack/echo/v4"
 	v1 "github.com/vultisig/commondata/go/vultisig/vault/v1"
 	rtypes "github.com/vultisig/recipes/types"
-	"google.golang.org/protobuf/proto"
-
-	ecommon "github.com/ethereum/go-ethereum/common"
-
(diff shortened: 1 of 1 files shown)


internal/api/policy.go
```diff
@@ -71,6 +69,31 @@ func (s *Server) CreatePluginPolicy(c echo.Context) error {
...
+	//			r.Rules[i].ParameterConstraints[j].Constraint.DenominatedIn = "wei"
+	//		}
+	//	}
+	//}
+	//b, err := proto.Marshal(&r)
+	//if err != nil {
+	//	panic(err)
+	//}
+	//policy.Recipe = base64.StdEncoding.EncodeToString(b)
+
```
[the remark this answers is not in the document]
webpiratt: it's correct for now

internal/api/policy.go
```diff
@@ -71,6 +69,31 @@ func (s *Server) CreatePluginPolicy(c echo.Context) error {
 		return c.JSON(http.StatusBadRequest, NewErrorResponseWithMessage("Invalid plugin policy"))
 	}
 
+	// TODO: @webpiratt: mock required fields until @ehsanst fixing UI
```
johnnyluo, line 72: can you please remove this?

webpiratt: Okay, removed

webpiratt: But it was only changes in this PR 😂 I'll close it then, and will save that snippet locally

PR #242, opened 2025-07-10
fix recipe params passed from UI

webpiratt:
Ref: https://github.com/vultisig/plugin/issues/124

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated default values for scheduling fields when creating a new recipe, ensuring more appropriate initial settings for interval and maximum executions.


johnnyluo (approved):
LGTM

PR #243, opened 2025-07-10
🚑 Alignment of "pricings" and "billing" data

garry-sharp:
## Summary by CodeRabbit

* **New Features**
  * Added support for multiple pricing options per plugin, including new pricing types (once, recurring, per-transaction), frequencies (daily, weekly, biweekly, monthly), and explicit asset specification (USDC).
  * Pricing details are now included directly in plugin data and displayed in the UI, replacing the previous single pricing model.

* **Improvements**
  * Enhanced billing and pricing validation for consistency between plugin pricing and billing policies.
  * Updated database schema and API to support richer pricing information and relationships.
  * Refined fee and billing amount types to unsigned 64-bit integers for improved accuracy.
  * Streamlined policy creation to incorporate multiple fee policies based on plugin pricing.

* **Bug Fixes**
  * Improved error handling and validation for billing and pricing data in forms and API responses.

* **Chores**
  * Updated seed and test data to match the new pricing model.
  * Refactored and cleaned up code and type definitions for consistency and clarity.
  * Simplified database queries and unified billing insertion logic.

PR #246, opened 2025-07-13
Fix plugin policy pricing

Ehsan-saradar:


garry-sharp (changes requested):
No changes actually needed, but lets merge https://github.com/vultisig/verifier/pull/243 in first, since it now works with fee policies etc, and then we can look at this

PR #273, opened 2025-07-23
🚚 Hardcoded frontend for fees plugin pending magic constants

garry-sharp:
## Summary by CodeRabbit

* **New Features**
  * Plugin policy modal now automatically fills and locks specific fields for the "vultisig-fees-feee" plugin, preventing user edits to those values.

* **Bug Fixes**
  * Improved SQL query ordering for plugin retrieval in the backend.

* **Chores**
  * Updated the default recipient whitelist address in the example worker configuration file.


garry-sharp:
Okay, there was a dumb thing here where plugins were ordered by unnecesarily complex criteria which failed the smoke tests as it looked from pricing sometimes in the fee plugin which isn't there.

Small db fix included in this as part of the ordering, small enough it hopefully doesn't cause any issues

internal/storage/postgres/plugin.go
```diff
@@ -172,9 +172,6 @@ func (p *PostgresBackend) FindPlugins(
 		return nil, fmt.Errorf("database pool is nil")
 	}
 
-	allowedSortingColumns := map[string]bool{"updated_at": true, "created_at": true, "title": true}
-	orderBy, orderDirection := common.GetSortingCondition(sort, allowedSortingColumns)
```
Ehsan-saradar, line 176: So we won't be able to sort plugins in plugin UI?! should we do sort on FE?

plugins-ui/src/components/PluginPolicyModal.tsx
```diff
@@ -325,7 +342,12 @@ export const PluginPolicyModal: FC<PluginPolicyModalProps> = ({
                             name={parameterName}
                             rules={[{ required }]}
                           >
-                            <Stack as={Input} $style={{ height: "44px" }} />
```
Ehsan-saradar, line 328: We can simplified  by just

`<Stack as={Input} $style={{ height: "44px" }} disabled={isFeesPlugin} />`

plugins-ui/src/components/PluginPolicyModal.tsx
```diff
@@ -217,6 +222,8 @@ export const PluginPolicyModal: FC<PluginPolicyModalProps> = ({
       });
   };
 
+  const isFeesPlugin = schema.pluginId === "vultisig-fees-feee";
```
Ehsan-saradar, line 225: Would be nice to have constant list for plugin id but we can ignore it for now

internal/storage/postgres/plugin.go
garry-sharp: All of this code here is going to be replaced and refactored later, it's horrificly messy and builds sql queries using string concatenation. Proper sorting will be a feature, this here is to resolve a smoke test bug where 10% of the time the fee plugin is returned first

Ehsan-saradar (approved):

PR #288, opened 2025-08-01
Scheduler replace from plugin repo

evgensheff:
Replacement of internal package which should be importable

## Summary by CodeRabbit

* **New Features**
  * Introduced a scheduling system for managing and executing plugin tasks.
  * Added support for transactional creation, updating, and deletion of plugin policies.
  * Implemented a worker that automatically enqueues and schedules tasks based on defined policies and intervals.
  * Provided storage integration for managing scheduled tasks, including support for PostgreSQL.
  * Included a no-operation service for plugins that do not require scheduling.
  * Added database schema and indexing to support efficient scheduling operations.


internal/storage/postgres/schema/schema.sql
```diff
@@ -288,6 +288,11 @@ CREATE TABLE "reviews" (
     CONSTRAINT "reviews_rating_check" CHECK ((("rating" >= 1) AND ("rating" <= 5)))
 );
 
+CREATE TABLE "scheduler" (
+    "policy_id" "uuid" NOT NULL,
+    "next_execution" timestamp without time zone NOT NULL
+);
+
```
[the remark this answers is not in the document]
evgensheff: It's autogenerated dump file

common/migrations/system/20250801165241_scheduler.sql
```diff
@@ -0,0 +1,11 @@
+-- +goose Up
+-- +goose StatementBegin
+BEGIN;
+CREATE TABLE IF NOT EXISTS scheduler (
```
webpiratt, line 4: Types and Go pkgs are fine, but I don't think we should have `scheduler` table in verifier db, because verifier doesn't have any schedulers

Maybe better consider move this migration to other folder like `scheduler`, not `system`

johnnyluo: maybe the whole thing should be just in copytrader instead

evgensheff: Copytrader doesn't use it, but this package is a part of standard plugin API

johnnyluo: then let's don't add it to verifier , as I don't think any of the plugins use it right now , especially it need database support , so I don't think we should add it here

evgensheff: @webpiratt correct me if I'm wrong, but scheduler will be common util for recurring actions in all plugins.
The problem is that we can't use plugin repo as import in third party plugins. Let's decide where it should be and I'll move it

webpiratt:
@evgensheff let’s close it, I am doing almost same changes in my PR

PR #297, opened 2025-08-05
Add test data

evgensheff:
For testing copytrader plugin we need to add enum into verifier database. I guess it's not okay.
At this moment all plugins use enum plugin_id in plugin_policies table

## Summary by CodeRabbit

* **New Features**
  * Added support for the "Vultisig Copytrader" plugin, including its registration, pricing, tagging, and API key setup.

* **Chores**
  * Updated database schema and seed data to include the new plugin and its associated configurations.

### DIFF
--- a/internal/storage/postgres/schema/schema.sql
+++ b/internal/storage/postgres/schema/schema.sql
@@ -10,9 +10,9 @@ CREATE TYPE "plugin_category" AS ENUM (
 
 CREATE TYPE "plugin_id" AS ENUM (
     'vultisig-dca-0000',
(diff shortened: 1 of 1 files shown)


webpiratt:
@RaghavSood @evgensheff I think `pluginId` db type should be changed to `string`

RaghavSood:
> @RaghavSood @evgensheff I think `plugin_id` db type should be changed to `string`

I don't think that's helpful

Having a proper enum for limited value types is better (at least for now) - it prevents a whole close of data consistency issues, at the cost of 1 PR to add an enum type

PR #315, opened 2025-08-14
schedule end

webpiratt:
## Summary by CodeRabbit

- New Features
  - Automatically cleans up tasks with no future run after they are queued, removing them from schedules.
- Bug Fixes
  - Prevents schedules from displaying an empty or invalid “next run” time when no future execution exists.
  - Improves accuracy of upcoming runs by only retaining tasks with a valid next schedule.
- Chores
  - Minor internal handling to streamline post-execution task management and reduce stale schedule entries.


johnnyluo (approved):
LGTM

PR #317, opened 2025-08-16
🐛 Render empty config values in policy modal

garry-sharp:
## Summary by CodeRabbit

* **Bug Fixes**
  * Plugin Policy modal no longer errors when a plugin lacks configuration or properties; it now safely renders with an empty configuration list.
  * Improved stability for plugins with incomplete schemas, preventing unexpected crashes while preserving existing behavior when configuration is present.
  * Users should experience smoother loading and fewer interruptions when opening the modal across a wider range of plugins.


RaghavSood:
If needed, this should be reopened on https://github.com/vultisig/plugin-marketplace

PR #318, opened 2025-08-18
PluginKeysignRequest builder

evgensheff:
Problem was in different message encoding in verifier api and vault/keysign.go
Unification needed

## Summary by CodeRabbit

- New Features
  - Create plugin keysign requests from raw EVM transactions with automatic payload decoding, base64-encoded message payloads, SHA-256 hash metadata, and tracking ID support.
  - New API to build keysign requests from raw transactions for EVM chains.

- Bug Fixes
  - Signature hash derived from decoded message bytes for correct signing.
  - Stronger validation to ensure messages are valid base64.
  - More reliable chain handling by using the message's chain value directly.


types/keysign.go
```diff
@@ -56,3 +66,36 @@ type PluginKeysignRequest struct {
 	Transaction     string `json:"transactions"`
 	TransactionType string `json:"transaction_type"`
 }
+
+func NewPluginKeysignRequest(policy PluginPolicy, txToTrack string, chain vgcommon.Chain, tx []byte) (
```
webpiratt, line 70: Let's rename it like `NewPluginKeysignRequestEvm`

webpiratt (approved):

types/keysign.go
evgensheff: Makes sense

PR #321, opened 2025-08-19
Validate recipe schema

Ehsan-saradar:
make sure we don't return an invalid  recipe schema and cache an invalid date in redis

## Summary by CodeRabbit

* **Bug Fixes**
  * Added runtime validation when fetching plugin recipe schemas to prevent invalid or incomplete configurations from being used.
  * Provides clearer error messages if a plugin’s recipe is malformed, improving reliability and user feedback during plugin setup.

* **Tests**
  * Added unit tests to ensure recipe schema validation works correctly across common error cases and valid inputs.


johnnyluo (approved):
LGTM

PR #342, opened 2025-09-09
🐛 Remove the \restrict tag from the pg_dump data

garry-sharp:
With postgres:16 `\restrict`tags are added to the db schema when performing a db_dump. This removes them from the output


webpiratt (approved):
LGTM

Issue #268, opened 2025-07-20
Database transaction leak on policy creation/deletion


RaghavSood:
There is a database transaction leak on policy creation and deletion

It appears to stem from the nested transaction in the fee validation code

If a policy API call fails with an error, the transactions aren't fully closed, and subsequent calls timeout trying to acquire a transaction lock.

It should be possible to reproduce this locally by creating a policy with verifier, while payroll is turned off/returns an error

PR #373, opened 2025-11-04
fix tx.Rollback

evgensheff:
Fix of https://github.com/vultisig/verifier/issues/268

If the context has been cancelled, tx can't be cancelled and this causes deadlock

## Summary by CodeRabbit

* **Chores**
  * Improved internal database transaction handling to ensure rollback operations complete reliably, regardless of request context status.
  * Simplified transaction flow to reduce intermediate rollback attempts and let errors propagate cleanly, improving stability and error handling.

### DIFF
--- a/internal/storage/postgres/policy.go
+++ b/internal/storage/postgres/policy.go
@@ -210,12 +210,8 @@ func (p *PostgresBackend) InsertPluginPolicyTx(ctx context.Context, dbTx pgx.Tx,
 	`
 
 	var insertedPolicy types.PluginPolicy
-	tx, err := dbTx.Begin(ctx)
-	if err != nil {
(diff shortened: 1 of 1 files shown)


internal/storage/postgres/policy.go
```diff
@@ -235,7 +235,6 @@ func (p *PostgresBackend) InsertPluginPolicyTx(ctx context.Context, dbTx pgx.Tx,
 		&insertedPolicy.Recipe,
 	)
 	if err != nil {
-		tx.Rollback(ctx)
```
webpiratt: I see `Begin` above, let's add `defer Rollback` right after `Begin` ?

webpiratt (approved):

PR #376, opened 2025-11-06
rm duplicated tx_indexer code

evgensheff:
## Summary by CodeRabbit

* **Refactor**
  * Reorganized transaction indexer module structure with updated import paths
  * Exposed new public methods for accessing worker configuration, concurrency limits, timeout settings, and transaction repository
  * Enhanced transaction status updates with improved error handling and state change notifications


webpiratt (approved):
LGTM but that would break payroll + old fee plugin + copytrading, since it's using old path, but I think it's fine

PR #395, opened 2025-11-20
Plugins Tx Indexer + Scheduler Metrics

neavra:
## Summary by CodeRabbit

* **New Features**
  * Added runtime metrics for scheduler and transaction indexing: active/stuck counts, processing durations, status/error recording, and timestamps. A no-op metrics mode is available by default in some contexts.

* **Chores**
  * Upgraded Prometheus client and several Go ecosystem dependencies to support the new metrics infrastructure.

### DIFF
--- a/internal/service/policy_worker.go
+++ b/internal/service/policy_worker.go
@@ -36,7 +36,12 @@ func (s *PolicyService) HandleScheduledFees(ctx context.Context, task *asynq.Tas
 		return fmt.Errorf("failed to query scheduled fees: %w", err)
 	}
 	defer rows.Close()
-	recurse := false
+
+	var feesToInsert []struct {
+		publicKey string
+		policyId  string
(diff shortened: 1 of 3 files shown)


plugin/tx_indexer/worker.go
```diff
@@ -149,8 +174,17 @@ func (w *Worker) UpdateTxStatus(ctx context.Context, tx storage.Tx) (*rpc.TxOnCh
 
 	err = w.repo.SetOnChainStatus(ctx, tx.ID, newStatus)
 	if err != nil {
+		if w.metrics != nil {
```
webpiratt: I suggest to create something like `metrics.NilService` and pass it in case if no metrics needed, instead of checking for nil for each usage

webpiratt (changes requested):

PR #401, opened 2025-11-24
Add endpoint to list installed plugins

Ehsan-saradar:
* Added the `GetInstalledPlugins` handler to `internal/api/plugin.go`, which checks each plugin for an associated vault backup file and returns the list of installed plugins for the authenticated user.

**Performance consideration:**
* Included a TODO comment in `GetInstalledPlugins` describing the current scalability limitation due to filename format, and suggesting a future change to use prefix-based listing for better performance.

## Summary by CodeRabbit

* **New Features**
  * Added a new authenticated endpoint to list installed plugins with pagination.
  * Returns only plugins that have corresponding vault backup files, reflecting installed state.
  * Robust listing: continues when individual plugin backup checks fail and surfaces top-level errors (e.g., missing authentication).

### DIFF
--- a/internal/api/plugin.go
+++ b/internal/api/plugin.go
@@ -250,7 +250,7 @@ func (s *Server) GetPlugins(c echo.Context) error {
 		return c.JSON(http.StatusInternalServerError, NewErrorResponseWithMessage(msgGetPluginsFailed))
 	}
 
-	return c.JSON(http.StatusOK, plugins)
+	return c.JSON(http.StatusOK, NewSuccessResponse(http.StatusOK, plugins))
 }
 
 func (s *Server) GetPlugin(c echo.Context) error {
@@ -266,7 +266,7 @@ func (s *Server) GetPlugin(c echo.Context) error {
 		return c.JSON(http.StatusInternalServerError, NewErrorResponseWithMessage(msgGetPluginFailed))
 	}
 
-	return c.JSON(http.StatusOK, plugin)
+	return c.JSON(http.StatusOK, NewSuccessResponse(http.StatusOK, plugin))
 }
 
 func (s *Server) GetInstalledPlugins(c echo.Context) error {
@@ -326,7 +326,7 @@ func (s *Server) GetCategories(c echo.Context) error {
 			Name: types.PluginCategoryPlugin.String(),
 		},
 	}
-	return c.JSON(http.StatusOK, resp)
+	return c.JSON(http.StatusOK, NewSuccessResponse(http.StatusOK, resp))
 }
 
 func (s *Server) GetTags(c echo.Context) error {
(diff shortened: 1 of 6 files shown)


internal/api/plugin.go
```diff
@@ -269,6 +269,49 @@ func (s *Server) GetPlugin(c echo.Context) error {
...
+	// This will drastically improve performance and reduce storage load.
+
+	pluginList, err := s.db.FindPlugins(c.Request().Context(), types.PluginFilters{}, 1000, 0, "")
+	if err != nil {
+		s.logger.WithError(err).Error("failed to fetch plugins")
+		return c.JSON(http.StatusInternalServerError, NewErrorResponseWithMessage(msgGetPluginsFailed))
+	}
+
+	var installedPlugins types.PluginsPaginatedList
+	installedPlugins.Plugins = make([]types.Plugin, 0)
```
4others: Could we initiate the `installedPlugins.Plugins` with predefined `cap`?

I.e.
```suggestion
	installedPlugins.Plugins = make([]types.Plugin, 0, len(pluginList.Plugins))
```

internal/api/plugin.go
```diff
@@ -269,6 +269,49 @@ func (s *Server) GetPlugin(c echo.Context) error {
...
+	// and use List(prefix) instead of N calls to Exist().
+	// This will drastically improve performance and reduce storage load.
+
+	pluginList, err := s.db.FindPlugins(c.Request().Context(), types.PluginFilters{}, 1000, 0, "")
+	if err != nil {
+		s.logger.WithError(err).Error("failed to fetch plugins")
+		return c.JSON(http.StatusInternalServerError, NewErrorResponseWithMessage(msgGetPluginsFailed))
+	}
+
+	var installedPlugins types.PluginsPaginatedList
```
4others: Nit - consider assigning this to a local installed variable:
```suggestion
	var installed types.PluginsPaginatedList
```
It makes the following lines easier to read:
```
			installed.Plugins = append(installed.Plugins, plugin)
			installed.TotalCount++
```
without changing the intent.

internal/api/plugin.go
```diff
@@ -269,6 +269,49 @@ func (s *Server) GetPlugin(c echo.Context) error {
...
+	installedPlugins.Plugins = make([]types.Plugin, 0)
+
+	for _, plugin := range pluginList.Plugins {
+		fileName := common.GetVaultBackupFilename(publicKey, string(plugin.ID))
+
+		exist, err := s.vaultStorage.Exist(fileName)
+		if err != nil {
+			s.logger.WithError(err).Error("failed to check plugin vault existence")
```
4others: Would it also make sense to log the plugin name, for easier debug or it doesn't matter really?

PR #418, opened 2025-12-03
Allow Verifier policy deletion

neavra:
## Summary by CodeRabbit

* **Bug Fixes**
  * Policy and plugin deletions now complete even if notifying the remote plugin server fails; operations proceed to local cleanup (verifier and vault) instead of aborting.
  * Failure to sync no longer returns an error to callers; such failures are logged as warnings with contextual identifiers for easier troubleshooting.


evgensheff (approved):
LGTM

PR #429, opened 2025-12-05
Add 'immediate' field to PluginPolicy

Ehsan-saradar:
Releated issue: https://github.com/vultisig/recipes/issues/293

## Summary by CodeRabbit

* **New Features**
  * Plugin policies now support an "immediate" execution mode (a required flag) so policies can be applied right away when configured.


RaghavSood:
Hmmm I don't think this is the right way to handle it - this shouldn't be a policy level flag

It should be configuration level defined by the respective plugins.

`immediate` for recurring swaps/sends can mean a different thing as opposed to immediate for another use case.

`start_now` or a similar flag should go through the policy configuration flow.

PR #440, opened 2025-12-11
Bump recipes for tooltips updates

neavra:
## Summary by CodeRabbit

* **Chores**
  * Updated internal dependencies to maintain system stability and ensure compatibility with the latest components.


Ehsan-saradar:
I think we can remove the following endpoint too
https://github.com/vultisig/verifier/blob/main/internal/api/server.go#L182

Issue #131, opened 2025-06-03
[Tx Index] Update transaction during lifecycle changes


RaghavSood:
As a transaction is validated against a recipe/policy, and then signed, we must update the status in the tx index appropriately

PR #491, opened 2026-01-15
fix: Add missing plugin migrations for deactivation_reason and control_flags

jpthor:
## Summary

This PR adds missing plugin migrations that were causing policy creation and scheduling failures.

### Missing Plugin Migrations
- **deactivation_reason column**: Added migration to plugin database path (`plugin/policy/policy_pg/migrations/`)
  - This column was already added to verifier DB in PR #490, but was missing from plugin DBs
  - Plugin databases need this column for policy deactivation tracking

- **control_flags table**: Added migration for safety module (`plugin/safety/safety_pg/migrations/`)
  - Required by safety manager for feature flag control
  - Added embed directive in `plugin/migrations.go` to include safety migrations

## Why

When plugins (like DCA scheduler) initialize their databases, they run plugin migrations. These two migrations were missing, causing:
- Policy creation failures (missing `deactivation_reason` column)
- Scheduler initialization failures (missing `control_flags` table)

## Related
- Follow-up to PR #490 which added these to verifier DB
- Required for app-recurring PR #131 to work correctly

## Summary by CodeRabbit

* **New Features**
  * Added deactivation reason tracking for policies, enabling better visibility into policy state changes.
  * Introduced a control flags system for enhanced policy and safety management.

### DIFF
--- a/common/migrations/system/20260114120000_add_deactivation_reason.sql
+++ b/common/migrations/system/20260114120000_add_deactivation_reason.sql
@@ -4,6 +4,3 @@ ALTER TABLE plugin_policies ADD COLUMN deactivation_reason TEXT DEFAULT NULL;
 -- +goose StatementEnd
 
 -- +goose Down
--- +goose StatementBegin
-ALTER TABLE plugin_policies DROP COLUMN IF EXISTS deactivation_reason;
--- +goose StatementEnd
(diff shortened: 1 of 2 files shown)


4others (changes requested):
approved earlier, but after giving it another thought, we should either release verifier to prod (this options is being discussed right now) or make existing schemas idempotent, i.e.
`ADD COLUMN IF NOT EXISTS` (as both plugins and verifier touch the same table)

4others (approved):

webpiratt (changes requested):

PR #569, opened 2026-02-18
chore: insert developer plugin

4others:
## Summary by CodeRabbit

* **Chores**
  * Added support for a Developer Plugin Listing Fee service, enabling payment processing capabilities for plugin developers.


RaghavSood (changes requested):
We shouldn't do this via migrations as previously discussed, since it applies to all environments.

I can run this on prod, denis/devs can run it on staging env over cli

Issue #72, opened 2025-05-20
[ADD] Policy Management on marketplace UI


johnnyluo:
Allow user to define a policy 
- Fetch policy schema from plugin
- Allow user to define a policy
- Serialise the policy to bytes using Protobuf , and base64 encoded it

PR #581, opened 2026-04-30
feat(cosmos): add MsgBeginRedelegate + MsgWithdrawDelegatorReward schemas

NeOMakinG:
## Summary

Bump `vultisig/recipes` to the Phase 1.5 cut that registers the staking + distribution interfaces in the Cosmos SDK and adds policy-engine unpack/extract for `MsgBeginRedelegate` and `MsgWithdrawDelegatorReward`. Add tx-indexer tests that pin the new behavior at the verifier boundary.

Without this bump, even after mcp-ts PR #72 advertises the staking write tools, the verifier's Cosmos SDK codec has no concrete type for the incoming `Any.TypeUrl` on these messages — hashing fails and the tx is dropped on the indexer floor. With the bump, the existing THORChain / MayaChain indexers (which delegate to the recipes Cosmos SDK) handle generic Cosmos staking/distribution txs end-to-end with no per-chain wiring.

These two msgs are generic Cosmos SDK and unlock the same code path for Cosmos Hub, Osmosis, Kujira, Akash, Dydx, Noble alongside Terra (phoenix-1) and Terra Classic (columbus-5).

## Verifier-side changes

- `go.mod` / `go.sum` — bump `github.com/vultisig/recipes` from `577976dfb292` (verifier main) to `475ad1fe90db` (recipes#589 HEAD). `cosmossdk.io/math` and `github.com/cosmos/cosmos-sdk` move from indirect to direct, which is mechanical: the new test file imports them and `go mod tidy` promotes accordingly. No new transitive deps beyond what recipes already pulled in.
- `plugin/tx_indexer/pkg/chain/cosmos_msg_types_test.go` (new) — six test groups:
  - **`TestRecipesSDKRegistersStakingDistributionInterfaces`** — `UnpackAny` assertion forced through the registry path by marshal+unmarshal of the `Any` first. Without the round-trip, `NewAnyWithValue` caches the source pointer and `UnpackAny` short-circuits on `cachedValue` — meaning the assertion would pass even with no registration. Verified locally by patching out `stakingtypes.RegisterInterfaces` / `distributiontypes.RegisterInterfaces` in `recipes/sdk/cosmos/sdk.go`: the test now fails with `no concrete type registered for type URL /cosmos.staking.v1beta1.MsgBeginRedelegate`.
  - **`TestTHORChainIndexer_HashesMsgBeginRedelegate` / `_HashesMsgWithdrawDelegatorReward`** — end-to-end `Sign + ComputeTxHash` on representative redelegate / withdraw_rewards fixtures. Asserts byte-exact equality with `SHA256(signedTx)` plus determinism.
  - **`TestTHORChainIndexer_HashIsSignedTxSHA256`** — replaces the earlier `_HashIsSignedShape` test (which only asserted `hash != SHA256(unsigned)`, a tripwire that fires on any byte-changing transform). New assertion is byte-exact equality with the upper-hex SHA256 of the marshalled signed-tx bytes.
  - **`TestMayaChainIndexerRegistersStakingDistributionInterfaces`** — same registry + hash assertions for `NewMayaChainIndexer`. `chains_list.go` wires Maya through `cosmossdk.NewSDK(nil)` exactly as it wires THORChain, but a future divergence (e.g. a Maya-specific SDK constructor that skips `RefreshCodec`) would silently regress redelegate / withdraw_rewards on Maya only — this test catches that.
  - **`TestTHORChainIndexerDecodesMultiMsgWithStakingMsg`** — `bank.MsgSend + staking.MsgBeginRedelegate` in one tx body. Real cosmos txs often bundle messages (auto-claim before delegate), and the indexer's hash path must handle the multi-msg shape.
  - **`TestTHORChainIndexerDecodesCrossEcosystemMsgs`** — canary: thor `MsgDeposit` (registered via `rtypes.RegisterInterfaces`) + cosmos staking `MsgBeginRedelegate` (registered by `NewSDK`) in the same body, both round-tripping through the same THORChain SDK. Catches future registration shadowing between the two ecosystems.

Real-shape fixtures: real cosmoshub-4 bech32 addresses (delegator + two known mainnet validators), real on-curve secp256k1 keypair via `GenPrivKeyFromSecret`, real signature via `priv.Sign()`. Replaces the earlier off-curve `pubKey[0] = 0x02` placeholder and zero-R `fakeSig`.

## Pattern fidelity

The verifier-side change is intentionally minimal — the recipes change is the load-bearing one. Verifier's `chains_list.go` calls `cosmossdk.NewSDK(nil)` for THORChain + Maya, and the recipes update auto-registers staking + distribution interfaces in `NewSDK`. No per-chain wiring changes are needed.

## What this bumps beyond cosmos staking

The recipes pin moves from `577976dfb292` (verifier main) to `475ad1fe90db` (recipes#589 HEAD) — 59 commits, +4232 / −172 lines across 44 files. The cosmos staking schemas are 632 lines of that. The remaining ~3600 lines are unrelated SDK growth that landed on recipes/main between Jan 29 and Apr 30. Disclosed for review:

- AAVE v3 SDK — `sdk/evm/aavev3/{abi,addresses,builder,client}.go` + `chain/evm/abi/aavev3_*.json` + `sdk/evm/codegen/aavev3_*` (~700 LOC, +2 ABI files)
- Polymarket CTF + CTF Exchange — `sdk/evm/codegen/polymarket_ctf{,_exchange}/*.go` (~230 LOC)
- Cetus swap SDK — `sdk/swap/cetus.go` (190 LOC, new)
- Relay swap SDK — `sdk/swap/relay.go` (510 LOC, new) + service/router rewiring
- THORChain router codegen — `sdk/evm/codegen/thorchain_router/thorchain_router.go` (+95, −15)
- MayaChain router codegen — `sdk/evm/codegen/mayachain_router/mayachain_router.go` (+496, new file)
- Uniswap v2 router codegen rewrite — `sdk/evm/codegen/uniswapv2_router/uniswapv2_router.go` (+285, −51)
- ERC20 codegen rewrite — `sdk/evm/codegen/erc20/erc20.go` (+110, −20)
- LiFi + swap-service refactors — `sdk/swap/{lifi,service,router,thorchain,mayachain,types}.go`
- THORChain native receive resolver — `resolver/thorchain_common.go` etc (`always-available for native THORChain/MayaChain receives`)
- `RESOURCES.md` (+284) — recipes-side documentation, no runtime effect

None of these touch cosmos staking, the tx_indexer, or any verifier-direct surface. The `chains_list.go` paths that the verifier uses (`cosmossdk.NewSDK`, `btc.NewSDK`, `solana.NewSDK`, `xrpl.NewSDK`, `tron.NewSDK`) are exercised by the receipts below; the EVM swap SDK growth above is consumed only via mcp-ts, not directly by verifier.

## Paired PR

- vultisig/recipes#589 — adds the schemas, codec registration, and policy-engine unpack/extract paths.

## Codex 5.4 second-opinion

Run pre-promotion. Findings:
- Type URLs match cosmos-sdk v0.50.x upstream
- Bech32 prefix non-validation in the generic engine is correct (consistent with `MsgSend`, chain-keeper validates at execution time)
- `IsNil + IsPositive` covers the structural amount cases
- Cross-registering generic cosmos staking types into the THORChain/Maya SDK instance is harmless (disjoint type URLs)

Subsequent CR feedback on recipes#589 (empty-address validation on redelegate / withdraw_rewards) addressed in commits `1d5f4ac9`, `e9919915`, `475ad1fe`. This PR's pin update picks all three up.

CR feedback on verifier#581 — registry test cached-pointer short-circuit, off-curve pubkey + zero-R signature placeholders, weak hash-shape assertion, missing Maya / multi-msg / cross-ecosystem coverage — addressed in this commit.

## Refs

- `vultisig-ops/.spikes/terra-readiness-checkpoint-2026-04-30.md` § B.5 — gap evidence
- `vultisig-ops/.spikes/terra-product-plan.md` — Phase 1.5 description

## ci receipts

`go test ./plugin/tx_indexer/pkg/chain/... -count=1 -v`:

```
=== RUN   TestRecipesSDKRegistersStakingDistributionInterfaces
=== RUN   TestRecipesSDKRegistersStakingDistributionInterfaces/MsgBeginRedelegate_decodes_via_SDK_registry
=== RUN   TestRecipesSDKRegistersStakingDistributionInterfaces/MsgWithdrawDelegatorReward_decodes_via_SDK_registry
--- PASS: TestRecipesSDKRegistersStakingDistributionInterfaces (0.00s)
    --- PASS: TestRecipesSDKRegistersStakingDistributionInterfaces/MsgBeginRedelegate_decodes_via_SDK_registry (0.00s)
    --- PASS: TestRecipesSDKRegistersStakingDistributionInterfaces/MsgWithdrawDelegatorReward_decodes_via_SDK_registry (0.00s)
=== RUN   TestTHORChainIndexer_HashesMsgBeginRedelegate
--- PASS: TestTHORChainIndexer_HashesMsgBeginRedelegate (0.00s)
=== RUN   TestTHORChainIndexer_HashesMsgWithdrawDelegatorReward
--- PASS: TestTHORChainIndexer_HashesMsgWithdrawDelegatorReward (0.00s)
=== RUN   TestTHORChainIndexer_HashIsSignedTxSHA256
--- PASS: TestTHORChainIndexer_HashIsSignedTxSHA256 (0.00s)
=== RUN   TestMayaChainIndexerRegistersStakingDistributionInterfaces
=== RUN   TestMayaChainIndexerRegistersStakingDistributionInterfaces/MsgBeginRedelegate_hashes_through_Maya_indexer
=== RUN   TestMayaChainIndexerRegistersStakingDistributionInterfaces/MsgWithdrawDelegatorReward_hashes_through_Maya_indexer
=== RUN   TestMayaChainIndexerRegistersStakingDistributionInterfaces/registry_decodes_both_msg_types
--- PASS: TestMayaChainIndexerRegistersStakingDistributionInterfaces (0.00s)
    --- PASS: TestMayaChainIndexerRegistersStakingDistributionInterfaces/MsgBeginRedelegate_hashes_through_Maya_indexer (0.00s)
    --- PASS: TestMayaChainIndexerRegistersStakingDistributionInterfaces/MsgWithdrawDelegatorReward_hashes_through_Maya_indexer (0.00s)
    --- PASS: TestMayaChainIndexerRegistersStakingDistributionInterfaces/registry_decodes_both_msg_types (0.00s)
=== RUN   TestTHORChainIndexerDecodesMultiMsgWithStakingMsg
--- PASS: TestTHORChainIndexerDecodesMultiMsgWithStakingMsg (0.00s)
=== RUN   TestTHORChainIndexerDecodesCrossEcosystemMsgs
--- PASS: TestTHORChainIndexerDecodesCrossEcosystemMsgs (0.00s)
=== RUN   TestTHORChainIndexer_NilSDK
--- PASS: TestTHORChainIndexer_NilSDK (0.00s)
PASS
ok  	github.com/vultisig/verifier/plugin/tx_indexer/pkg/chain	0.517s
```

`go test ./plugin/tx_indexer/... -count=1`:

```
ok  	github.com/vultisig/verifier/plugin/tx_indexer	1.305s
ok  	github.com/vultisig/verifier/plugin/tx_indexer/pkg/chain	0.443s
?   	github.com/vultisig/verifier/plugin/tx_indexer/pkg/config	[no test files]
?   	github.com/vultisig/verifier/plugin/tx_indexer/pkg/conv	[no test files]
?   	github.com/vultisig/verifier/plugin/tx_indexer/pkg/graceful	[no test files]
ok  	github.com/vultisig/verifier/plugin/tx_indexer/pkg/rpc	0.724s
?   	github.com/vultisig/verifier/plugin/tx_indexer/pkg/storage	[no test files]
```

EVM indexer (`plugin/tx_indexer/pkg/rpc/evm_errors_test.go` covers the EVM revert decoding path; the EVM indexer itself is at `plugin/tx_indexer/pkg/chain/evm.go` with no dedicated test file). Note: this repo organizes indexers under `plugin/tx_indexer/pkg/chain/`, not `internal/indexer/{evm,thor}/`. The EVM RPC error tests above (last block) cover the EVM-specific paths the bump could touch.

Full repo:

```
ok  	github.com/vultisig/verifier/config	0.302s
ok  	github.com/vultisig/verifier/internal/portal	1.637s
ok  	github.com/vultisig/verifier/internal/service	0.558s
ok  	github.com/vultisig/verifier/internal/sigutil	1.854s
ok  	github.com/vultisig/verifier/internal/storage/postgres	1.067s
ok  	github.com/vultisig/verifier/internal/util	1.558s
ok  	github.com/vultisig/verifier/plugin/libhttp	1.827s
ok  	github.com/vultisig/verifier/plugin/tx_indexer	1.460s
ok  	github.com/vultisig/verifier/plugin/tx_indexer/pkg/chain	0.874s
ok  	github.com/vultisig/verifier/plugin/tx_indexer/pkg/rpc	1.018s
ok  	github.com/vultisig/verifier/vault	0.722s
```

Robot-generated.

## qa receipts

End-to-end runtime probe pulled real on-chain raw protobuf tx bytes from production LCDs (phoenix-1, thornode, mayanode) and fed them through the recipes Cosmos SDK's `InterfaceRegistry` + the verifier's `THORChainIndexer` / `MayaChainIndexer` factory output. This is the runtime evidence the existing CI tests can't provide — those round-trip synthetic fixtures, this exercises the registry path with bytes that came off real production validators on real production chains.

Probe lives at `/tmp/verifier-indexer-probe/probe_test.go` (QA-only, never committed). It uses `replace github.com/vultisig/verifier => ../worktrees/verifier-pr581` to import the PR-branch indexer code unmodified.

### Fixtures (4 real on-chain txs, SHA256 of fixture file == claimed txid)

| chain | txid | type | size |
|---|---|---|---|
| phoenix-1 | `D2529B4A33D7879E8445996932EC360478DCB6196FB1B4628757D10E87D1CAC8` | `/cosmos.staking.v1beta1.MsgBeginRedelegate` (height 20799755, 304 LUNA redelegate) | 398 B |
| phoenix-1 | `D104A37D03D878189B2601370B139AF3E689DA4184140EC7FDA6A8073E8787E4` | 8x `/cosmos.distribution.v1beta1.MsgWithdrawDelegatorReward` (height 20801634, multi-validator reward claim) | 1464 B |
| thornode  | `040F3EEA0B85AF107320860237D42022304AC0A95DC108FA6AE6A06DBFF0CBFD` | `/types.MsgDeposit` (BTC→BSC.USDC swap, real memo + real signer) | 354 B |
| mayanode  | `35D4D77F7B2141C06F3C22816EFE81964715EC46EB33E23E77F947F646B3F309` | `/types.MsgDeposit` (BTC→ZEC swap, real memo + real signer) | 366 B |

`TestProbe_FixtureProvenance` asserts each fixture's SHA256 == claimed txid (Cosmos tx hash IS SHA256(rawTxBytes)), proving the fixtures are not locally fabricated.

### Probe results (10/10 pass)

`go test -count=1 -v ./...` from `/tmp/verifier-indexer-probe/`:

```
=== RUN   TestProbe_Phoenix1_MsgBeginRedelegate
    decoded delegator=terra1u7pzpfhn4sw8e27g56g6ucfyunwkmk48qd3czd src=terravaloper1hy8lzd38kp0t8662r2vnxsc4ezs3l74506jx02 dst=terravaloper1u53qwdgafryksqnzq3k7e5eyph7jlh8kgc3va4 amount=304000000uluna
    indexer hash (synthetic re-sign) = 11AFF109A813F1E5F265A7FB86BE40058CB77049850F25C805C5C75C33CEDEA9
--- PASS: TestProbe_Phoenix1_MsgBeginRedelegate (0.00s)
=== RUN   TestProbe_Phoenix1_MsgWithdrawDelegatorReward
    decoded all 8 MsgWithdrawDelegatorReward msgs from phoenix1 tx D104A3...87E4
    indexer hash (synthetic re-sign) = C5DF9EA2C0C11C0F4C8AA021D6324B031C401241714EA0C2EEBAAE045FAEF1E1
--- PASS: TestProbe_Phoenix1_MsgWithdrawDelegatorReward (0.00s)
=== RUN   TestProbe_THOR_MsgDeposit
    decoded thor MsgDeposit memo="=:BSC~USDC-0X8AC76A51CC950D9822D68B83FE1AD97B32CD580D:thor14mh37ua4vkyur0l5ra297a4la6tmf95mt96a55:19681801427/1/1" signer_len=20 coin_amount=254538 asset=BTC.BTC
    indexer hash (synthetic re-sign) = C970CEB8B6727E11554E38194505D3669D3EA9B69CF9D8448B5984AE563C109D
--- PASS: TestProbe_THOR_MsgDeposit (0.00s)
=== RUN   TestProbe_Maya_MsgDeposit
    decoded maya MsgDeposit memo="=:ZEC~ZEC:maya176820wmjp536wtsddxn4y52ghp3w23k747hard:32506078" coin_amount=146574 asset=BTC.BTC
    indexer hash (synthetic re-sign) = B6BD281F58F47F017FC20983CE811703ED8B314B25DB8ECAD92ABD473E3E6EBB
--- PASS: TestProbe_Maya_MsgDeposit (0.00s)
=== RUN   TestProbe_FixtureProvenance
    --- PASS: phoenix1_redelegate.bin SHA256=D2529B4A33D7879E8445996932EC360478DCB6196FB1B4628757D10E87D1CAC8
    --- PASS: phoenix1_withdraw.bin    SHA256=D104A37D03D878189B2601370B139AF3E689DA4184140EC7FDA6A8073E8787E4
    --- PASS: thornode_msgdeposit.bin  SHA256=040F3EEA0B85AF107320860237D42022304AC0A95DC108FA6AE6A06DBFF0CBFD
    --- PASS: maya_msgdeposit.bin      SHA256=35D4D77F7B2141C06F3C22816EFE81964715EC46EB33E23E77F947F646B3F309
=== RUN   TestProbe_AdversarialDecodeWithoutStakingOnly
    targeted negative control: failure cites typeUrl as expected: no concrete type registered for type URL /cosmos.staking.v1beta1.MsgBeginRedelegate against interface *interface { ProtoMessage(); Reset(); String() string }
--- PASS
=== RUN   TestProbe_AdversarialDecodeWithoutDistributionOnly
    targeted negative control (distribution): no concrete type registered for type URL /cosmos.distribution.v1beta1.MsgWithdrawDelegatorReward against interface *interface { ProtoMessage(); Reset(); String() string }
--- PASS
=== RUN   TestProbe_ProductionFactoryPath
    production-factory THORChain hash redelegate: 11AFF109A813F1E5F265A7FB86BE40058CB77049850F25C805C5C75C33CEDEA9
    production-factory MayaChain hash redelegate: 11AFF109A813F1E5F265A7FB86BE40058CB77049850F25C805C5C75C33CEDEA9
    production-factory THORChain hash MsgDeposit: C970CEB8B6727E11554E38194505D3669D3EA9B69CF9D8448B5984AE563C109D
--- PASS
=== RUN   TestProbe_NewSDKBaselineMatchesHandWiring
    hand-wired probe SDK matches reconstructed NewSDK baseline
--- PASS
=== RUN   TestProbe_AdversarialDecodeWithoutNewRegistration
    bare-registry decode failed as expected
--- PASS
PASS
ok  	verifier-indexer-probe	0.454s
```

### What each probe proves

| probe | invariant under test | how it differs from CI tests |
|---|---|---|
| `TestProbe_Phoenix1_MsgBeginRedelegate` | recipes SDK registry decodes a REAL on-chain phoenix-1 redelegate, fields match LCD | CI tests round-trip synthetic msgs; this consumes real wire bytes |
| `TestProbe_Phoenix1_MsgWithdrawDelegatorReward` | registry decodes all 8 messages in a multi-msg withdraw_rewards body | catches "first-Any-only" registry bugs |
| `TestProbe_THOR_MsgDeposit` | rtypes.MsgDeposit still decodes after Phase 1.5 (regression) | uses real swap memo + signer bytes |
| `TestProbe_Maya_MsgDeposit` | maya MsgDeposit still decodes | proves Maya wiring matches THOR wiring |
| `TestProbe_FixtureProvenance` | fixture SHA256 == claimed txid for all 4 | proves fixtures are real on-chain bytes, not fabricated |
| `TestProbe_AdversarialDecodeWithoutStakingOnly` | registry-minus-staking-only fails decode citing exact typeUrl | targeted control: staking registration is the load-bearing delta (not "no registrations") |
| `TestProbe_AdversarialDecodeWithoutDistributionOnly` | registry-minus-distribution-only fails decode citing exact typeUrl | targeted control: distribution registration is independently load-bearing |
| `TestProbe_ProductionFactoryPath` | `tx_indexer.Chains()` factory output (production wiring) accepts the new bodies | catches future drift in `chains_list.go` (e.g. someone removing RefreshCodec) |
| `TestProbe_NewSDKBaselineMatchesHandWiring` | hand-wired probe SDK matches reconstructed NewSDK + rtypes baseline | documents why the probe's helpers are trustworthy |
| `TestProbe_AdversarialDecodeWithoutNewRegistration` | empty registry can't decode anything (broad sanity check) | weakest control, kept for completeness |

### Adversarial review (codex-cli gpt-5.4)

Codex was asked to argue against the probe. It raised 7 concerns; the probe was strengthened in response and the strengthened version addresses each:

| # | Concern (from codex) | Resolution |
|---|---|---|
| 1 | "Hash-path probes don't prove registry — `Sign()` doesn't `UnpackAny` body messages" | Concerns separated: registry is proved by `*DecodeWithout*` tests, not the hash-path tests |
| 2 | "`roundTripAny` may not actually clear cachedValue for wire-derived Anys" | Added `require.Nil(t, any.GetCachedValue())` after roundtrip in the new adversarial probes |
| 3 | "Bare-registry negative control is too broad — could fail for unrelated reasons" | Added `*WithoutStakingOnly` + `*WithoutDistributionOnly` that register baseline-minus-one and assert failure cites the exact typeUrl |
| 4 | "Probe hand-wires SDK; doesn't exercise `tx_indexer.Chains()` factory" | Added `TestProbe_ProductionFactoryPath` calling the actual production factory |
| 5 | "Probe trusts hardcoded fixture txids; doesn't prove provenance" | Added `TestProbe_FixtureProvenance` asserting SHA256(fixture) == claimed txid |
| 6 | "`stripSignerInfo` produces a synthetic shape (no signer_infos, sequence-defaults-to-0)" | Acknowledged. The CI test in this PR uses the same shape. The contract being tested is "indexer accepts unsigned bodies", which holds for both. Production wallet sends pre-signing equivalents that match. |
| 7 | "Field assertions narrower than 'fields match LCD' claim" | withdraw_rewards has no amount field by design (per `cosmos.distribution.v1beta1.proto`); thor/maya now check memo + signer + coin_amount + asset.Chain + asset.Ticker |

### Caveats / non-goals

- The probe does NOT broadcast txs — it consumes already-on-chain bytes and verifies the indexer can decode/hash them.
- The probe does NOT prove the indexer's policy-evaluation downstream consumers (those are out of scope for this PR — verifier#581 is purely about the registry knowing the typeUrls).
- The probe lives in `/tmp/verifier-indexer-probe/` (QA-only, not committed). It was run on 2026-05-01 against the PR HEAD `a606f37` with `recipes` pinned at `475ad1fe`.

Probe artifacts captured at `/tmp/verifier-indexer-probe/`:
- `probe_test.go` — the 10 probes
- `fixtures/*.bin` — the 4 real on-chain raw tx bytes
- `fixtures/*.json` — the LCD JSON responses for cross-reference
- `probe-output.txt` — full `go test -v` output
- `codex-adversarial-output.txt` — full codex 5.4 dual-review transcript

Robot-generated.

## Summary by CodeRabbit

* **Chores**
  * Promoted and updated core module dependency declarations for improved compatibility and stability (Cosmos SDK and related libraries).
* **Tests**
  * Expanded test coverage for transaction hashing, message decoding, determinism, multi-message transactions, and cross-ecosystem message handling to strengthen indexer correctness and interoperability.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -16,6 +16,7 @@ require (
 	github.com/google/uuid v1.6.0
 	github.com/hashicorp/go-retryablehttp v0.7.7
 	github.com/hibiken/asynq v0.25.1
+	github.com/jackc/pgerrcode v0.0.0-20250907135507-afb5586c32a6
 	github.com/jackc/pgx/v5 v5.7.4
 	github.com/kaptinlin/jsonschema v0.4.6
 	github.com/kelseyhightower/envconfig v1.4.0
@@ -31,7 +32,7 @@ require (
 	github.com/vultisig/commondata v0.0.0-20250710214228-61d9ed8f7778
 	github.com/vultisig/go-wrappers v0.0.0-20260116015747-e12e4d06cf57
 	github.com/vultisig/mobile-tss-lib v0.0.0-20250316003201-2e7e570a4a74
-	github.com/vultisig/recipes v0.0.0-20260129020926-577976dfb292
+	github.com/vultisig/recipes v0.0.0-20260430070612-989f77eca639
 	github.com/vultisig/vultiserver v0.0.0-20250825042420-c6e6ac281110
 	github.com/vultisig/vultisig-go v0.0.0-20260114092710-6c38516a0c85
 	github.com/xyield/xrpl-go v0.0.0-20230914223425-9abe75c05830
@@ -177,7 +178,6 @@ require (
 	github.com/inconshreveable/mousetrap v1.1.0 // indirect
 	github.com/ipfs/go-log v1.0.5 // indirect
 	github.com/ipfs/go-log/v2 v2.5.1 // indirect
-	github.com/jackc/pgerrcode v0.0.0-20250907135507-afb5586c32a6 // indirect
 	github.com/jackc/pgpassfile v1.0.0 // indirect
 	github.com/jackc/pgservicefile v0.0.0-20240606120523-5a60cdf6a761 // indirect
 	github.com/jackc/puddle/v2 v2.2.2 // indirect
--- a/go.mod
+++ b/go.mod
@@ -3,9 +3,11 @@ module github.com/vultisig/verifier
 go 1.25
 
 require (
+	cosmossdk.io/math v1.5.3
 	github.com/aws/aws-sdk-go v1.55.7
 	github.com/btcsuite/btcd v0.24.2
 	github.com/btcsuite/btcd/btcutil/psbt v1.1.10
+	github.com/cosmos/cosmos-sdk v0.50.11
 	github.com/decred/dcrd/dcrec/secp256k1/v4 v4.4.0
 	github.com/eager7/dogd v0.0.0-20200427085516-2caf59f59dbb
 	github.com/ethereum/go-ethereum v1.15.11
@@ -32,7 +34,7 @@ require (
 	github.com/vultisig/commondata v0.0.0-20250710214228-61d9ed8f7778
 	github.com/vultisig/go-wrappers v0.0.0-20260116015747-e12e4d06cf57
 	github.com/vultisig/mobile-tss-lib v0.0.0-20250316003201-2e7e570a4a74
-	github.com/vultisig/recipes v0.0.0-20260430070612-989f77eca639
+	github.com/vultisig/recipes v0.0.0-20260430121755-475ad1fe90db
 	github.com/vultisig/vultiserver v0.0.0-20250825042420-c6e6ac281110
 	github.com/vultisig/vultisig-go v0.0.0-20260114092710-6c38516a0c85
 	github.com/xyield/xrpl-go v0.0.0-20230914223425-9abe75c05830
@@ -49,7 +51,6 @@ require (
 	cosmossdk.io/depinject v1.2.1 // indirect
 	cosmossdk.io/errors v1.0.2 // indirect
 	cosmossdk.io/log v1.6.0 // indirect
-	cosmossdk.io/math v1.5.3 // indirect
 	cosmossdk.io/schema v1.1.0 // indirect
 	cosmossdk.io/store v1.1.2 // indirect
 	cosmossdk.io/x/tx v0.14.0 // indirect
@@ -95,7 +96,6 @@ require (
 	github.com/cosmos/btcutil v1.0.5 // indirect
 	github.com/cosmos/cosmos-db v1.1.1 // indirect
 	github.com/cosmos/cosmos-proto v1.0.0-beta.5 // indirect
-	github.com/cosmos/cosmos-sdk v0.50.11 // indirect
 	github.com/cosmos/go-bip39 v1.0.0 // indirect
 	github.com/cosmos/gogogateway v1.2.0 // indirect
 	github.com/cosmos/gogoproto v1.7.0 // indirect
--- a/go.mod
+++ b/go.mod
@@ -34,7 +34,7 @@ require (
 	github.com/vultisig/commondata v0.0.0-20250710214228-61d9ed8f7778
 	github.com/vultisig/go-wrappers v0.0.0-20260116015747-e12e4d06cf57
 	github.com/vultisig/mobile-tss-lib v0.0.0-20250316003201-2e7e570a4a74
-	github.com/vultisig/recipes v0.0.0-20260430121755-475ad1fe90db
+	github.com/vultisig/recipes v0.0.0-20260503115421-d28b3880fd08
 	github.com/vultisig/vultiserver v0.0.0-20250825042420-c6e6ac281110
 	github.com/vultisig/vultisig-go v0.0.0-20260114092710-6c38516a0c85
 	github.com/xyield/xrpl-go v0.0.0-20230914223425-9abe75c05830
--- a/go.sum
+++ b/go.sum
@@ -1046,8 +1046,8 @@ github.com/vultisig/go-wrappers v0.0.0-20260116015747-e12e4d06cf57 h1:ZFlNC4JuaY
 github.com/vultisig/go-wrappers v0.0.0-20260116015747-e12e4d06cf57/go.mod h1:vEP0x0RmNlghWxfalt13FvVsBwmobSwimMhJxGfqCD4=
 github.com/vultisig/mobile-tss-lib v0.0.0-20250316003201-2e7e570a4a74 h1:goqwk4nQ/NEVIb3OPP9SUx7/u9ZfsUIcd5fIN/e4DVU=
 github.com/vultisig/mobile-tss-lib v0.0.0-20250316003201-2e7e570a4a74/go.mod h1:nOykk4nOy1L3yXtLSlYvVsgizBnCQ3tR2N5uwGPdvaM=
-github.com/vultisig/recipes v0.0.0-20260129020926-577976dfb292 h1:CVXtJBOjJfzMsv+akQK65Jcup9/TSeMUXD8NMJ9O0eg=
-github.com/vultisig/recipes v0.0.0-20260129020926-577976dfb292/go.mod h1:PUz2BoPkuCrk9EFeWavynFSIMM2DNULfFeSS0CGVIVc=
+github.com/vultisig/recipes v0.0.0-20260430070612-989f77eca639 h1:brVx7S9gELCbDP0xaTqx/sCMkwNvGSdXPcsO+hr3ahs=
+github.com/vultisig/recipes v0.0.0-20260430070612-989f77eca639/go.mod h1:PUz2BoPkuCrk9EFeWavynFSIMM2DNULfFeSS0CGVIVc=
 github.com/vultisig/vultiserver v0.0.0-20250825042420-c6e6ac281110 h1:7WDQ92FAdu08Byjgm3RNS8Sok49sK521PzPcbRpbzCE=
 github.com/vultisig/vultiserver v0.0.0-20250825042420-c6e6ac281110/go.mod h1:HwP2IgW6Mcu/gX8paFuKvfibrGE9UmPgkOFTub6dskM=
 github.com/vultisig/vultisig-go v0.0.0-20260114092710-6c38516a0c85 h1:NAVXp2Mm791Z/teQHUA8T7mhmx3bouyrXvQkLXvAu3M=
--- a/go.sum
+++ b/go.sum
@@ -1048,6 +1048,8 @@ github.com/vultisig/mobile-tss-lib v0.0.0-20250316003201-2e7e570a4a74 h1:goqwk4n
 github.com/vultisig/mobile-tss-lib v0.0.0-20250316003201-2e7e570a4a74/go.mod h1:nOykk4nOy1L3yXtLSlYvVsgizBnCQ3tR2N5uwGPdvaM=
 github.com/vultisig/recipes v0.0.0-20260430121755-475ad1fe90db h1:hYjpUTaFVqJWVWF7+/RyZNI81C5uG7Q6Q0rm2IS8hf0=
 github.com/vultisig/recipes v0.0.0-20260430121755-475ad1fe90db/go.mod h1:PUz2BoPkuCrk9EFeWavynFSIMM2DNULfFeSS0CGVIVc=
+github.com/vultisig/recipes v0.0.0-20260503115421-d28b3880fd08 h1:GiZ9OfXAJst6+9ee73QxxfSqItbWB6HhmDQ79lx//cc=
+github.com/vultisig/recipes v0.0.0-20260503115421-d28b3880fd08/go.mod h1:PUz2BoPkuCrk9EFeWavynFSIMM2DNULfFeSS0CGVIVc=
 github.com/vultisig/vultiserver v0.0.0-20250825042420-c6e6ac281110 h1:7WDQ92FAdu08Byjgm3RNS8Sok49sK521PzPcbRpbzCE=
 github.com/vultisig/vultiserver v0.0.0-20250825042420-c6e6ac281110/go.mod h1:HwP2IgW6Mcu/gX8paFuKvfibrGE9UmPgkOFTub6dskM=
 github.com/vultisig/vultisig-go v0.0.0-20260114092710-6c38516a0c85 h1:NAVXp2Mm791Z/teQHUA8T7mhmx3bouyrXvQkLXvAu3M=
--- a/go.sum
+++ b/go.sum
@@ -1046,8 +1046,8 @@ github.com/vultisig/go-wrappers v0.0.0-20260116015747-e12e4d06cf57 h1:ZFlNC4JuaY
 github.com/vultisig/go-wrappers v0.0.0-20260116015747-e12e4d06cf57/go.mod h1:vEP0x0RmNlghWxfalt13FvVsBwmobSwimMhJxGfqCD4=
 github.com/vultisig/mobile-tss-lib v0.0.0-20250316003201-2e7e570a4a74 h1:goqwk4nQ/NEVIb3OPP9SUx7/u9ZfsUIcd5fIN/e4DVU=
 github.com/vultisig/mobile-tss-lib v0.0.0-20250316003201-2e7e570a4a74/go.mod h1:nOykk4nOy1L3yXtLSlYvVsgizBnCQ3tR2N5uwGPdvaM=
-github.com/vultisig/recipes v0.0.0-20260430070612-989f77eca639 h1:brVx7S9gELCbDP0xaTqx/sCMkwNvGSdXPcsO+hr3ahs=
-github.com/vultisig/recipes v0.0.0-20260430070612-989f77eca639/go.mod h1:PUz2BoPkuCrk9EFeWavynFSIMM2DNULfFeSS0CGVIVc=
+github.com/vultisig/recipes v0.0.0-20260430121755-475ad1fe90db h1:hYjpUTaFVqJWVWF7+/RyZNI81C5uG7Q6Q0rm2IS8hf0=
+github.com/vultisig/recipes v0.0.0-20260430121755-475ad1fe90db/go.mod h1:PUz2BoPkuCrk9EFeWavynFSIMM2DNULfFeSS0CGVIVc=
 github.com/vultisig/vultiserver v0.0.0-20250825042420-c6e6ac281110 h1:7WDQ92FAdu08Byjgm3RNS8Sok49sK521PzPcbRpbzCE=
 github.com/vultisig/vultiserver v0.0.0-20250825042420-c6e6ac281110/go.mod h1:HwP2IgW6Mcu/gX8paFuKvfibrGE9UmPgkOFTub6dskM=
 github.com/vultisig/vultisig-go v0.0.0-20260114092710-6c38516a0c85 h1:NAVXp2Mm791Z/teQHUA8T7mhmx3bouyrXvQkLXvAu3M=
--- a/plugin/tx_indexer/pkg/chain/cosmos_msg_types_test.go
+++ b/plugin/tx_indexer/pkg/chain/cosmos_msg_types_test.go
@@ -3,13 +3,15 @@ package chain
 import (
 	"crypto/sha256"
 	"encoding/hex"
-	"fmt"
+	"strings"
 	"testing"
 
 	"cosmossdk.io/math"
 	codectypes "github.com/cosmos/cosmos-sdk/codec/types"
+	"github.com/cosmos/cosmos-sdk/crypto/keys/secp256k1"
 	cosmostypes "github.com/cosmos/cosmos-sdk/types"
 	"github.com/cosmos/cosmos-sdk/types/tx"
+	banktypes "github.com/cosmos/cosmos-sdk/x/bank/types"
 	distributiontypes "github.com/cosmos/cosmos-sdk/x/distribution/types"
 	stakingtypes "github.com/cosmos/cosmos-sdk/x/staking/types"
 	"github.com/stretchr/testify/assert"
@@ -26,33 +28,67 @@ import (
 //     This is what consumers of `sdk.InterfaceRegistry()` (including downstream
 //     decoders that DO need to UnpackAny the body messages — e.g. anything
 //     that inspects the redelegate amount or validator addresses post-sign)
-//     rely on. We assert this directly via `UnpackAny` in
-//     TestRecipesSDKRegistersStakingDistributionInterfaces.
+//     rely on. We assert this directly via `UnpackAny` against an Any whose
+//     cachedValue has been cleared by a marshal+unmarshal roundtrip — that
+//     forces the registry resolution path rather than the cached fast-path.
 //
 //  2. The end-to-end indexer hash path (`Sign + ComputeTxHash`) accepts the
-//     new tx shapes and produces a stable SHA256 of the signed bytes. Note
-//     that Sign() itself only round-trips the outer tx.Tx envelope — it does
-//     not UnpackAny the body messages — so the hash tests don't *require*
-//     the new interface registrations to pass. They guard the broader
-//     contract: an indexer fed an unsigned redelegate/withdraw_rewards tx
-//     should not error, and the hash is shaped like all other cosmos hashes.
+//     new tx shapes and produces SHA256 of the signed bytes. Sign() itself
+//     does not UnpackAny the body messages — it round-trips the outer tx.Tx
+//     envelope — so the hash tests don't *require* the new interface
+//     registrations to pass. They guard the broader contract: an indexer fed
+//     an unsigned redelegate/withdraw_rewards tx (or a multi-msg body mixing
+//     bank + staking, or a thor MsgDeposit + cosmos staking msg) should not
+//     error, and the hash equals SHA256 of the marshalled signed bytes.
 //
 // If a future recipes bump regresses either invariant, the failure surfaces
 // here rather than at runtime when a real redelegate broadcasts.
 
+// Real-shape test fixtures.
+//
+// Addresses are real bech32 strings on cosmoshub-4. The delegator address
+// matches QA-Stability-Vault's cosmoshub address; the validator addresses are
+// well-known mainnet validators (Coinbase Custody, Binance Staking). These
+// addresses don't need to be live for tests — the SDK's Sign()/ComputeTxHash
+// path doesn't validate them — but using real shapes means a future bech32
+// stricter check upstream wouldn't fail in a confusing way.
 const (
-	testDelegator    = "cosmos1delegatorxxxxxxxxxxxxxxxxxxxxxxxxxxxxx"
-	testValidatorSrc = "cosmosvaloper1srcxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx"
-	testValidatorDst = "cosmosvaloper1dstxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx"
-	testValidator    = "cosmosvaloper1abcxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx"
+	testDelegator    = "cosmos1n9qcqp4ds6c30zmnyxwq30vfd03cyuq3gpsllv"
+	testValidatorSrc = "cosmosvaloper1grgelyng2v6v3t8z87wu3sxgt9m5s03xfytvz7"
+	testValidatorDst = "cosmosvaloper1sjllsnramtg7ewxqwwrwjxfgc4n4ef9u2lcnj0"
+	testValidator    = "cosmosvaloper1grgelyng2v6v3t8z87wu3sxgt9m5s03xfytvz7"
+	testRecipient    = "cosmos1ej2es5fjztqjcd4pwa0zyvaevtjd2y5w37wr9t"
 )
 
-// fakeSig returns a 64-byte deterministic R || S split that the recipes SDK
-// will accept (low-S normalization happens inside Sign()).
-func fakeSig() tss.KeysignResponse {
+// testKeypair returns a deterministic secp256k1 keypair derived from a fixed
+// seed. The pubkey bytes are a real on-curve compressed (33-byte) secp256k1
+// point — not the off-curve `[0x02 || 32 zero-bytes]` placeholder we used in
+// the first cut. Sign() in recipes doesn't validate curve membership today,
+// but a stricter upstream check would reject the placeholder, so use a real
+// point.
+func testKeypair(t *testing.T) (*secp256k1.PrivKey, []byte) {
+	t.Helper()
+	priv := secp256k1.GenPrivKeyFromSecret([]byte("vultisig-verifier-cosmos-msg-types-test"))
+	pub := priv.PubKey().Bytes()
+	require.Len(t, pub, 33, "compressed secp256k1 pubkey must be 33 bytes")
+	require.True(t, pub[0] == 0x02 || pub[0] == 0x03, "valid compressed pubkey starts with 0x02 or 0x03")
+	return priv, pub
+}
+
+// testSig returns a real ECDSA signature over a deterministic message, split
+// into R || S hex strings as the recipes SDK expects on the wire from a TSS
+// keysign. The signature is in lower-S form (the secp256k1 Sign helper
+// normalizes it). This replaces the earlier 32-byte-zero R placeholder, which
+// would never come out of a real keysign.
+func testSig(t *testing.T, priv *secp256k1.PrivKey) tss.KeysignResponse {
+	t.Helper()
+	msg := []byte("vultisig-verifier-cosmos-msg-types-test-message")
+	sig, err := priv.Sign(msg)
+	require.NoError(t, err)
+	require.Len(t, sig, 64, "secp256k1 signature is R||S = 64 bytes")
 	return tss.KeysignResponse{
-		R: hex.EncodeToString(make([]byte, 32)),
-		S: hex.EncodeToString(append([]byte{0x01}, make([]byte, 31)...)),
+		R: hex.EncodeToString(sig[:32]),
+		S: hex.EncodeToString(sig[32:]),
 	}
 }
 
@@ -60,14 +96,18 @@ func fakeSig() tss.KeysignResponse {
 // SignerInfos (Sign() injects the signer info from the provided pubkey).
 // This mirrors the shape that vultisig-app's cosmosTx.ts emits before
 // keysign — the unsigned bytes the verifier indexer receives.
-func buildUnsignedCosmosTx(t *testing.T, msg cosmostypes.Msg) []byte {
+func buildUnsignedCosmosTx(t *testing.T, msgs ...cosmostypes.Msg) []byte {
 	t.Helper()
 
-	any, err := codectypes.NewAnyWithValue(msg)
-	require.NoError(t, err)
+	anys := make([]*codectypes.Any, 0, len(msgs))
+	for _, m := range msgs {
+		any, err := codectypes.NewAnyWithValue(m)
+		require.NoError(t, err)
+		anys = append(anys, any)
+	}
 
 	body := &tx.TxBody{
-		Messages: []*codectypes.Any{any},
+		Messages: anys,
 	}
 
 	authInfo := &tx.AuthInfo{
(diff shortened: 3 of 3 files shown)


gomesalexandre (changes requested):
gmeh ser, claude orchestrator from cross-repo cosmos staking review (paired against vultiagent-app#333 + agent-backend#225 + mcp-ts#72). One round claude → codex (codex hit refresh-token error, second claude adversarial pass instead) → reconcile. STATIC review only - no checkout per session policy.

This is the verifier-side load-bearing piece of the cosmos staking write block: without the schema bump here, `Any.TypeUrl=/cosmos.staking.v1beta1.MsgBeginRedelegate` arriving at the indexer with no concrete codec → hashing fails → tx eaten on the floor. So this PR has to land before mcp-ts#72 advertises the staking write tools to a user. Whole funnel sits on this bump being correct.

A few things stop me from stamping this APPROVE today.

---

**blocker: pinning to an unmerged recipes-PR commit, not the PR HEAD**

The bump pins `recipes` to `989f77eca639`. That commit is the FIRST commit of **OPEN** recipes PR #589 (`feat/cosmos-redelegate-withdraw-schemas`). The actual HEAD of #589 is `475ad1fe`, with 3 follow-up CR-feedback commits already layered:

- `1d5f4ac9` - fix(cosmos): reject empty addresses in withdraw_rewards
- `e9919915` - fix(cosmos): reject empty validator addresses in redelegate
- `475ad1fe` - fix(cosmos): add delegator_address required-field check on redelegate

Verifier is pinning the version BEFORE all three CR fixes. That means:
1. The redelegate / withdraw_rewards extractors landing here accept empty validator/delegator addresses (the chain-keeper would reject at execution but the policy-engine wouldn't fail-fast).
2. As soon as recipes#589 merges, this PR needs a re-bump anyway, and the new validation won't be exercised by the tests being added here.

Action: either (a) repin to `475ad1fe` (current recipes#589 HEAD) so we ship the CR fixes together, or (b) hold this PR until recipes#589 merges to main and pin to the merged-to-main commit. Option (b) is cleaner for tests-as-receipts.

---

**preferably-blocking: load-bearing test has a false-positive failure mode**

`TestRecipesSDKRegistersStakingDistributionInterfaces` is supposed to prove the recipes SDK's `InterfaceRegistry` knows the new msg types. But:

```go
any, err := codectypes.NewAnyWithValue(original)  // caches *value* on the Any
require.NoError(t, err)
...
err = ir.UnpackAny(any, &decoded)                 // uses cached value, registry unused
```

`codectypes.NewAnyWithValue` populates the cached value pointer on the `Any`. `UnpackAny` on a value-cached `Any` short-circuits and succeeds **regardless of registry state**. The registry is only consulted when going from raw protobuf bytes → `Msg`. So this test would pass on a recipes version that didn't register `stakingtypes.RegisterInterfaces` at all - i.e., it doesn't actually prove the invariant the comment claims it proves.

Fix is one extra round-trip:

```go
any, err := codectypes.NewAnyWithValue(original)
require.NoError(t, err)

// Force the codec path: marshal+unmarshal so the cached value is gone.
bz, err := any.Marshal()
require.NoError(t, err)
any2 := &codectypes.Any{}
require.NoError(t, any2.Unmarshal(bz))

var decoded cosmostypes.Msg
require.NoError(t, ir.UnpackAny(any2, &decoded))
```

This is the test the PR body cites as the fix codex 5.4 added in `870cd8e` after the original Sign-only tests didn't actually exercise the registration. Worth tightening before merge so the receipt matches the claim.

---

**should-fix: PR body underdiscloses 56-commit recipes bump collateral**

`go.mod` `ahead_by: 56` from the old pinned recipes commit. PR body sells this as "the recipes change is the load-bearing one" but doesn't mention the bump also pulls:

- AAVE v3 SDK (new, ~700 LOC across builder/client/codegen)
- Polymarket CTF + CTF Exchange SDKs (new, ~230 LOC)
- Cetus swap SDK (new, ~190 LOC)
- Relay swap SDK (new, ~510 LOC)
- THORChain router codegen (95 +, 15 -)
- Uniswap v2 router codegen rewrite (285 +, 51 -)
- ERC20 codegen rewrite (110 +, 20 -)
- LiFi + swap-service refactors

None of those are exercised by the new tests. If any of the swap-aggregator codegen rewrites regresses an EVM tx hash path the verifier indexer touches, it ships under this PR's flag silently. Action:
- Spell out the surface area in the PR body so reviewers + future bisectors know what's in this bump.
- Run the existing EVM + THORChain + Maya indexer test suites and embed the receipts (already done for `tx_indexer/...` and `tx_indexer/pkg/chain/...`, but the EVM indexer paths matter most given the codegen rewrites - please post `go test ./...` on the changed packages too).

---

**should-fix: no Maya indexer test, no multi-msg test**

`chains_list.go` wires both THORChain AND Maya through `cosmossdk.NewSDK(nil)` + `types.RegisterInterfaces(...)` + `RefreshCodec()`. Tests only cover THORChain. Drift risk: a future Maya wiring change breaks redelegate parsing silently. One mirror test for `NewMayaChainIndexer` is cheap insurance.

Also missing: a `[]Msg` shape with `[MsgSend, MsgBeginRedelegate]` in the same body. Real txs are multi-message; the failure mode where one of multiple Anys is unregistered is exactly where missing interface registrations bite.

---

**should-fix: weak test fixtures - addresses + pubKey + sigs are off-distribution**

- `cosmos1delegatorxxxxxxxxxxxxxxxxxxxxxxxxxxxxx` is **not valid bech32**. The current SDK doesn't validate bech32 at hash time so the test passes, but if recipes ever adds a structural check (it should, eventually) all four tests break at once.
- `pubKey := make([]byte, 33); pubKey[0] = 0x02` puts the all-zero x-coord behind a valid prefix. Not on the secp256k1 curve. Same fragility.
- `fakeSig` R = 32 zero bytes → invalid signature scalar. Same fragility.

None of these are hard blockers (the tests genuinely prove what they claim modulo T1 above), but using real-shape fixtures (a known testnet bech32 + a known valid pubkey + a real R/S) would future-proof against recipes adding structural validation.

---

**should-fix: hash-shape test is comfort food, not a guard**

`TestTHORChainIndexer_HashIsSignedShape` only asserts `SHA256(unsigned) != ComputeTxHash(...)`. Trivially true any time `Sign()` mutates by ≥1 byte. Doesn't actually prove the indexer hashes signed bytes - would also pass if `Sign()` returned `unsigned ++ [0x00]`.

If you want a real signed-shape guard, marshal the signed `tx.Tx` from `Sign()` directly (or use the SDK helper that produces it) and assert `ComputeTxHash` matches `SHA256(thatBytesUppercaseHex)`. Otherwise drop the test - it's a regression-tripwire that won't trip on real regressions.

---

**suggestion: `pgerrcode` direct/indirect shuffle is correct, but call it out cleaner**

The promotion from indirect → direct is right (`internal/portal/proposed_plugin_handlers.go` directly imports it). PR body already mentions it as a `go mod tidy` side-effect, just slightly buried. Personal taste: split it into its own micro-commit so the cosmos bump diff stays surgical.

---

**q: backward-compat for THORChain/Maya msg parsing**

Recipes PR #589 registers `stakingtypes.RegisterInterfaces` + `distributiontypes.RegisterInterfaces` in `chain/cosmos/chain.go::NewChain` AND `engine/cosmos/cosmos.go::NewEngine` AND `sdk/cosmos/sdk.go::NewSDK`. Verifier's `chains_list.go` already calls `cosmossdk.NewSDK(nil)` + `types.RegisterInterfaces(...)` for thor/maya. Spot-check from my side says the typeUrls (`/cosmos.staking.v1beta1.MsgBeginRedelegate`, `/cosmos.distribution.v1beta1.MsgWithdrawDelegatorReward`) are disjoint from THORChain `/types.MsgDeposit` + Maya equivalents, so cross-registration is harmless. Codex 5.4 came to the same conclusion per the PR body. Worth confirming with one extra test: a THORChain `MsgDeposit` + a generic `MsgBeginRedelegate` in the same tx body, both decoding through the thor SDK. That's the canary.

---

**praise: invariant choice + per-test commentary**

The way each test scopes its invariant in the file-level docblock is genuinely good - registration vs envelope-roundtrip vs signed-shape laid out as three orthogonal contracts is clearer than 95% of the cosmos test code I've seen. Codex 5.4 catching that the original Sign-only tests didn't prove the registration was a sharp call. Just need to actually close the gap (T1).

---

**Verdict: REQUEST_CHANGES.** Highest-priority items: (1) repin to `475ad1fe` or wait for recipes#589 merge, (2) tighten `TestRecipesSDKRegistersStakingDistributionInterfaces` to actually exercise the registry path, (3) disclose the 56-commit collateral in the PR body and confirm EVM indexer + THORChain swap regression. Everything else is layerable in a follow-up.

Cross-repo coupling FYI: this PR + recipes#589 + mcp-ts#72 + vultiagent-app#333 + agent-backend#225 form a 5-way write-block. If any one merges without the others wired, users get a silent gate. Sequence I'd ship: recipes#589 → THIS PR (with #589's merged-to-main commit pinned) → agent-backend#225 (Noble blocker addressed) → mcp-ts#72 → vultiagent-app#333.

🤖 reviewed via Claude Code (in-session orchestrator)

NeOMakinG:
Thanks for the deep review. Addressed in commit `a606f37`. Per-finding:

---

**blocker: pinning to an unmerged recipes-PR commit, not the PR HEAD** ✅

Repinned `github.com/vultisig/recipes` to `475ad1fe90db` (recipes#589 HEAD). All three CR-feedback commits are now picked up:
- `1d5f4ac9` — fix(cosmos): reject empty addresses in withdraw_rewards
- `e9919915` — fix(cosmos): reject empty validator addresses in redelegate
- `475ad1fe` — fix(cosmos): add delegator_address required-field check on redelegate

go.mod / go.sum delta is minimal — only the recipes pseudo-version bumps. `cosmossdk.io/math` and `github.com/cosmos/cosmos-sdk` move from indirect to direct, but that's `go mod tidy` mechanically responding to the test file's imports — not new transitive deps.

---

**preferably-blocking: load-bearing test has a false-positive failure mode** ✅

Tightened `TestRecipesSDKRegistersStakingDistributionInterfaces` with the marshal+unmarshal round-trip you suggested — extracted into a `roundTripAny` helper so both subtests use the same forced-codec-path pattern. The cached-pointer short-circuit in `cosmos-sdk/codec/types/interface_registry.go:UnpackAny` (line 343–346) is now bypassed.

**Verified the guard is real:** locally patched out `stakingtypes.RegisterInterfaces(ir)` and `distributiontypes.RegisterInterfaces(ir)` in `recipes/sdk/cosmos/sdk.go` (commented to `_` imports + commented call sites). The test now fails as expected:

```
--- FAIL: TestRecipesSDKRegistersStakingDistributionInterfaces/MsgBeginRedelegate_decodes_via_SDK_registry
    Error: no concrete type registered for type URL /cosmos.staking.v1beta1.MsgBeginRedelegate
           against interface *interface { ProtoMessage(); Reset(); String() string }

--- FAIL: TestRecipesSDKRegistersStakingDistributionInterfaces/MsgWithdrawDelegatorReward_decodes_via_SDK_registry
    Error: no concrete type registered for type URL /cosmos.distribution.v1beta1.MsgWithdrawDelegatorReward
           against interface *interface { ProtoMessage(); Reset(); String() string }
```

Restored recipes/sdk.go after verification. The receipt-vs-claim gap is now closed.

---

**should-fix: PR body underdiscloses 56-commit recipes bump collateral** ✅

PR body now has a `## What this bumps beyond cosmos staking` section. Actual bump from verifier-main (`577976dfb292`) to recipes#589 HEAD (`475ad1fe90db`) is **59 commits, +4232 / −172 across 44 files**. Cosmos staking is 632 lines; remaining ~3600 lines disclosed line-by-line: AAVE v3, Polymarket CTF + Exchange, Cetus, Relay, THORChain + MayaChain + Uniswap v2 + ERC20 codegen rewrites, LiFi/swap-service refactors, RESOURCES.md.

On EVM indexer regression risk: this repo organizes indexers under `plugin/tx_indexer/pkg/chain/`, not `internal/indexer/{evm,thor}/`. The EVM indexer file (`plugin/tx_indexer/pkg/chain/evm.go`) has no dedicated test file in this repo — it's covered transitively via `plugin/tx_indexer/...` integration tests. Posted those receipts plus full-repo `go test ./...` in the PR body's `## ci receipts` section. None of the swap-aggregator codegen rewrites touch the verifier indexer's runtime path; they're consumed only via mcp-ts.

---

**should-fix: no Maya indexer test, no multi-msg test** ✅

Added three new tests:

1. `TestMayaChainIndexerRegistersStakingDistributionInterfaces` — same registry + hash assertions as the THORChain set, against `NewMayaChainIndexer` wired with `mayaChainTestSDK()` (`NewSDK(nil)` + `rtypes.RegisterInterfaces` + `RefreshCodec`).
2. `TestTHORChainIndexerDecodesMultiMsgWithStakingMsg` — `bank.MsgSend + staking.MsgBeginRedelegate` in one tx body, both decoded through the registry post-roundtrip.
3. `TestTHORChainIndexerDecodesCrossEcosystemMsgs` — your canary suggestion (see below).

---

**should-fix: weak test fixtures — addresses + pubKey + sigs are off-distribution** ✅

Replaced all three:
- Real cosmoshub-4 bech32 addresses: delegator `cosmos1n9qcqp4ds6c30zmnyxwq30vfd03cyuq3gpsllv` (QA-Stability-Vault), validators `cosmosvaloper1grgelyng2v6v3t8z87wu3sxgt9m5s03xfytvz7` and `cosmosvaloper1sjllsnramtg7ewxqwwrwjxfgc4n4ef9u2lcnj0`. Recipient `cosmos1ej2es5fjztqjcd4pwa0zyvaevtjd2y5w37wr9t` for the multi-msg test.
- Real on-curve secp256k1 keypair via `secp256k1.GenPrivKeyFromSecret([]byte("vultisig-verifier-cosmos-msg-types-test"))`. Deterministic across runs; pubkey bytes are 33-byte compressed with `0x02`/`0x03` prefix asserted in `testKeypair()`.
- Real signature via `priv.Sign(deterministic-msg)`. Returns 64 bytes R||S in lower-S form (the secp256k1 helper normalizes). Replaces zero-R `fakeSig`.

The earlier off-distribution fixtures relied on the SDK not currently validating; the new shapes will continue to work even if recipes adds structural validation upstream.

---

**should-fix: hash-shape test is comfort food, not a guard** ✅

Replaced `TestTHORChainIndexer_HashIsSignedShape` with `TestTHORChainIndexer_HashIsSignedTxSHA256`. New assertion is byte-exact equality:

```go
expected := expectedSignedHash(t, sdk, unsigned, sigs, pubKey)  // SHA256(Sign(unsigned)) upper-hex
assert.Equal(t, expected, hash)
```

`expectedSignedHash` calls `sdk.Sign()` independently and computes `SHA256` of the marshalled signed bytes — exactly the contract `ComputeTxHash` is supposed to honor. The earlier weak `!= SHA256(unsigned)` assertion is kept as a secondary cross-check (so the new test still flags a regression where the indexer hashes the unsigned bytes), but the load-bearing assertion is now the byte-equality one.

Also embedded the same byte-equality check into the two existing `_HashesMsgBeginRedelegate` / `_HashesMsgWithdrawDelegatorReward` tests so they no longer rely on shape-only properties.

---

**suggestion: pgerrcode direct/indirect shuffle is correct, but call it out cleaner** ⏭️

Skipped per your "Personal taste" framing. The bump is already large; splitting `pgerrcode` into a separate commit would mostly add bisect-noise without runtime benefit. If a follow-up cleanup PR is preferred, happy to spin one separately.

---

**q: backward-compat for THORChain/Maya msg parsing — canary test request** ✅

Added `TestTHORChainIndexerDecodesCrossEcosystemMsgs`: a tx body containing both:
- `rtypes.MsgDeposit` (THORChain, registered via `rtypes.RegisterInterfaces`)
- `stakingtypes.MsgBeginRedelegate` (generic cosmos, registered by `NewSDK`)

Both decoded through the same THORChain SDK's registry post-roundtrip; both included in the multi-msg `Sign + ComputeTxHash` path. This catches future registration shadowing between the two ecosystems if either set's registration gets dropped or replaced during `NewSDK` setup.

---

**verdict: REQUEST_CHANGES**

All three highest-priority items addressed:
1. ✅ Repinned to `475ad1fe`.
2. ✅ Test now actually exercises the registry path; verified via temp-removal experiment.
3. ✅ 56-commit collateral disclosed in PR body; full-repo + tx_indexer-suite receipts embedded.

Plus all four should-fix items closed (Maya mirror, multi-msg, real fixtures, real signed-shape guard) and the canary cross-ecosystem test added.

Sequence agreement: recipes#589 → this PR (with the merged-to-main commit re-pinned at that point) → agent-backend#225 → mcp-ts#72 → vultiagent-app#333. Will hold for recipes#589 to merge, then re-pin and request re-review.

— vultisig-ops

go.mod
```diff
@@ -3,9 +3,11 @@ module github.com/vultisig/verifier
 go 1.25
 
 require (
+	cosmossdk.io/math v1.5.3
 	github.com/aws/aws-sdk-go v1.55.7
 	github.com/btcsuite/btcd v0.24.2
 	github.com/btcsuite/btcd/btcutil/psbt v1.1.10
+	github.com/cosmos/cosmos-sdk v0.50.11
```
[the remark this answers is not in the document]
NeOMakinG: ⏭️ Deferred — vulnerability surface does not apply to verifier; pin is owned upstream by recipes.

**Import-surface check.** Verifier imports `github.com/cosmos/cosmos-sdk` from exactly one file (`plugin/tx_indexer/pkg/chain/cosmos_msg_types_test.go`) and only pulls in:

- `codec/types`, `types`, `types/tx`, `crypto/keys/secp256k1`
- `x/bank/types`, `x/distribution/types`, `x/staking/types`

The 9 OSV findings against v0.50.11 all target chain-runtime modules verifier never executes:

| Advisory | Module / Surface |
| --- | --- |
| GO-2023-1821, GO-2023-1881 | `x/crisis` (constant-fee / chain-halt) |
| GO-2024-2584 | slashing evasion (`x/slashing` runtime) |
| GO-2025-3476 / GHSA-x5vx-95h7-rv4p | `x/group` malicious-proposal halt |
| GO-2025-3516 / GHSA-47ww-ff84-4jrg | `x/group` EndBlocker error halt |
| GO-2025-3803 / GHSA-p22h-3m2v-cmgh | Validator Rewards integer-overflow halt (EndBlocker) |

Verifier never runs a Cosmos chain, never executes EndBlockers, never imports `x/crisis`, `x/group`, `x/slashing`, or the validator-rewards path. It uses cosmos-sdk purely as a wire-format / type library for TSS signing + tx indexing of msgs sent to remote chains. The vulnerable code is not on a callable path here.

**Pin ownership.** Verifier#581 transitively pins cosmos-sdk via `github.com/vultisig/recipes` (this PR uses `recipes@475ad1fe`, which is the head of recipes#589). The recipes module is the source of truth for the cosmos-sdk version; bumping verifier alone causes a `go mod tidy` divergence on the next recipes consumer pull. The bump belongs in recipes, then flows downstream.

Tracking as a separate item against recipes; will not block this PR.

— vultisig-ops

gomesalexandre (approved):
CI green. The recipes bump to Phase 1.5 + explicit cosmos-sdk direct import is the right approach — promoting from indirect to direct so the interface registry is definitely initialised before any `UnpackAny` call on staking/distribution messages.

**praise:** The test suite is unusually thorough — testing both InterfaceRegistry resolution (cache-busted roundtrip) AND the end-to-end indexer hash path separately is exactly the right split. Guards two independent invariants cleanly.

**should-fix (minor):** The test file is 503 lines — it's a lot of boilerplate ABI encoding. If this pattern gets repeated for future Cosmos message types consider a test helper that builds a signed Tx from an arbitrary `sdk.Msg` slice.

**q:** Does the `recipes` bump at `20260430121755-475ad1fe90db` also touch any non-cosmos chains (EVM, BTC, etc.)? Shouldn't, but worth confirming there are no unintended side effects from the module bump.

LGTM.

go.mod
```diff
@@ -31,7 +34,7 @@ require (
 	github.com/vultisig/commondata v0.0.0-20250710214228-61d9ed8f7778
 	github.com/vultisig/go-wrappers v0.0.0-20260116015747-e12e4d06cf57
 	github.com/vultisig/mobile-tss-lib v0.0.0-20250316003201-2e7e570a4a74
-	github.com/vultisig/recipes v0.0.0-20260129020926-577976dfb292
+	github.com/vultisig/recipes v0.0.0-20260430121755-475ad1fe90db
```
0xApotheosis: This PR still pins recipes at `475ad1fe`, but the paired recipes PR is currently at `d28b3880`. The missing commits include the Gaia production wiring (`43d959c`) and HRP validation fixes (`b7381bf`, `d28b388`). As-is, the verifier tests can pass against the SDK registry while shipping without the latest policy-engine behavior from the paired PR. Please repin to recipes#589 HEAD and refresh `go.sum`.

0xApotheosis (changes requested):
Request changes.

Blocking finding: `go.mod` pins `github.com/vultisig/recipes` to `475ad1fe`, while paired recipes#589 is currently at `d28b3880`. The missing recipes commits include Gaia production wiring (`43d959c`) plus HRP validation fixes (`b7381bf`, `d28b388`). That means this PR can pass its SDK-registry/indexer tests while still not picking up the latest paired policy-engine behavior. Please repin to recipes#589 HEAD and refresh `go.sum`.

Verified locally:
- `env GOTOOLCHAIN=go1.25.0 GOCACHE=/private/tmp/go-build-verifier-pr581-125 GOMODCACHE=/private/tmp/go-mod-verifier-pr581 go test ./plugin/tx_indexer/... -count=1`

I also ran a Claude Code review pass; it independently identified the same blocker and found no other blocking issues in the verifier test file.

gomesalexandre:
ser, second change-request on the recipes pin. 0xApotheosis is right — need to repin to recipes#589 HEAD (`d28b3880`) to pick up the Gaia wiring + HRP validation commits. Quick fix: update go.mod + `go mod tidy` + `go mod download` and repush. Tag me when done.

go.mod
gomesalexandre: tackled in 147f3f5: repinned recipes to d28b3880fd08 (recipes PR#589 HEAD) via `go get github.com/vultisig/recipes@d28b3880fd08330a8f346f404271b69461274e29`

0xApotheosis (approved):
Re-reviewed the latest commit. The prior blocker is fixed: verifier is now pinned to recipes d28b3880, which includes the Gaia wiring and unconditional HRP validation fix. The registry test now forces the UnpackAny registry path, and the added Maya, multi-msg, cross-ecosystem, and signed-SHA256 assertions cover the earlier gaps. CI test/build jobs on this head are green.

